PR #34, opened 2020-03-26
fixes #33

### LINKED ISSUE #33: don't force fetching GetLastModified on AddressBook queries
I noticed that icloud doesn't add the GetLastModified property on `addressbook-multiget` and `addrssbook-query` calls, however `decodeAddressList()`[see code here](https://github.com/emersion/go-webdav/blob/master/carddav/client.go#L239) returns error when not exists.

AlmogBaku:


emersion:
This breaks the case where the server returns a 404 for the prop.

I think it would make sense to remove `missingPropError` and replace it with:

```go
var errPropMissing = HTTPErrorf(http.StatusNotFound, "property missing from response")
```

AlmogBaku:
@emersion I guess that should do it :)

AlmogBaku:
hmm... I think your point make more sense. I'll try finding some time doing it.
right now im researching about the Sync implementation... (hopefully to do a PR)

PR #36, opened 2020-03-28
fixes #33, remove missingPropError error

### LINKED ISSUE #33: don't force fetching GetLastModified on AddressBook queries

AlmogBaku:
Replaces #34

### DIFF
--- a/carddav/server.go
+++ b/carddav/server.go
@@ -124,7 +124,7 @@ func (h *Handler) handleQuery(w http.ResponseWriter, query *addressbookQuery) er
 	var q AddressBookQuery
 	if query.Prop != nil {
 		var addressData addressDataReq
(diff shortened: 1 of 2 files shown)


AlmogBaku:
instead of #34

AlmogBaku:
@emersion

emersion:
I get notifications for all pull requests. Please wait a week before pinging me, my review queue is rarely empty so it's normal if I don't give feedback immediately.

Thanks for the updated pull request, I'll try to get to it soon.

Issue #35, opened 2020-03-28
Supporting "Collection Synchronization" operations


AlmogBaku:
Low-level support for Collection Synchronizations as described in [This RFC](https://tools.ietf.org/html/rfc6578)

PR #37, opened 2020-03-29
sync-collection for client

AlmogBaku:
Basic sync-collection operation for client
partial implementation of #35

### DIFF
--- a/carddav/carddav.go
+++ b/carddav/carddav.go
@@ -99,3 +99,17 @@ type AddressObject struct {
 	ETag    string
 	Card    vcard.Card
 }
+
+//SyncQuery is the query struct represents a sync-collection request
+type SyncQuery struct {
+	DataRequest AddressDataRequest
+	SyncToken   string
+	Limit       int // <= 0 means unlimited
+}
+
+//SyncResponse contains the returned sync-token for next time
+type SyncResponse struct {
+	SyncToken string
+	Updated   []AddressObject
+	Deleted   []string
+}
--- a/carddav/client.go
+++ b/carddav/client.go
@@ -429,3 +429,56 @@ func (c *Client) PutAddressObject(path string, card vcard.Card) (*AddressObject,
 	}
 	return ao, nil
 }
+
+// SyncCollection do a sync-collection operation on resource(path), it returns a SyncResponse
+func (c *Client) SyncCollection(path string, query *SyncQuery) (*SyncResponse, error) {
+	var limit *internal.Limit
+	if query.Limit > 0 {
+		limit = &internal.Limit{NResults: uint(query.Limit)}
+	}
+
+	propReq, err := encodeAddressPropReq(&query.DataRequest)
+	if err != nil {
+		return nil, err
+	}
+
+	ms, err := c.ic.SyncCollection(path, query.SyncToken, internal.DepthOne, limit, propReq)
+	if err != nil {
+		return nil, err
+	}
+
+	ret := &SyncResponse{SyncToken: ms.SyncToken}
+	for _, resp := range ms.Responses {
+		p, err := resp.Path()
+		if err != nil {
+			if err, ok := err.(*internal.HTTPError); ok && err.Code == http.StatusNotFound {
+				ret.Deleted = append(ret.Deleted, p)
+				continue
+			}
+			return nil, err
+		}
+
+		if p == path || path == fmt.Sprintf("%s/", p) {
+			continue
+		}
+
+		var getLastMod internal.GetLastModified
+		if err := resp.DecodeProp(&getLastMod); err != nil && !internal.IsNotFound(err) {
+			return nil, err
+		}
+
+		var getETag internal.GetETag
+		if err := resp.DecodeProp(&getETag); err != nil && !internal.IsNotFound(err) {
+			return nil, err
+		}
+
+		o := AddressObject{
+			Path:    p,
+			ModTime: time.Time(getLastMod.LastModified),
+			ETag:    string(getETag.ETag),
+		}
+		ret.Updated = append(ret.Updated, o)
+	}
+
+	return ret, nil
+}
--- a/internal/client.go
+++ b/internal/client.go
@@ -174,3 +174,25 @@ func (c *Client) Options(path string) (classes map[string]bool, methods map[stri
 	methods = parseCommaSeparatedSet(resp.Header["Allow"], true)
 	return classes, methods, nil
 }
+
+// SyncCollection perform a `sync-collection` REPORT operation on a resource
+func (c *Client) SyncCollection(path, syncToken string, level Depth, limit *Limit, prop *Prop) (*Multistatus, error) {
+	q := SyncCollectionQuery{
+		SyncToken: syncToken,
+		SyncLevel: string(level),
+		Limit:     limit,
+		Prop:      prop,
+	}
+
+	req, err := c.NewXMLRequest("REPORT", path, &q)
+	if err != nil {
+		return nil, err
+	}
+
+	ms, err := c.DoMultiStatus(req)
+	if err != nil {
+		return nil, err
+	}
+
+	return ms, nil
+}
--- a/internal/elements.go
+++ b/internal/elements.go
@@ -93,6 +93,7 @@ type Multistatus struct {
 	XMLName             xml.Name   `xml:"DAV: multistatus"`
 	Responses           []Response `xml:"response"`
 	ResponseDescription string     `xml:"responsedescription,omitempty"`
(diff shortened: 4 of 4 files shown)


client.go
```diff
@@ -242,3 +242,32 @@ func (c *Client) MoveAll(name, dest string, overwrite bool) error {
...
+
+type SyncLevel string
+
+const (
+	SyncLevel1        SyncLevel = "1"
+	SyncLevelInfinite SyncLevel = "infinite"
+)
+
+// SyncCollection perform a `sync-collection` REPORT operation on a resource
+func (c *Client) SyncCollection(resource, syncToken string, level SyncLevel, limit *internal.Limit, prop *internal.Prop) (*internal.Multistatus, error) {
```
emersion, line 254: We can't expose `internal.Prop` like this.

internal/xml_test.go
```diff
@@ -56,7 +56,7 @@ func TestRawXMLValue_TokenReader(t *testing.T) {
 		if err == io.EOF {
 			break
 		} else if err != nil {
-			t.Fatalf("TokenReader.Token() = %v", err)
+			t.Fatalf("TokenReader.SyncToken() = %v", err)
```
emersion, line 59: Global find-and-replace replaced too much?

internal/elements.go
```diff
@@ -404,3 +405,18 @@ type Set struct {
...
 	Prop    Prop     `xml:"prop"`
 }
+
+// https://tools.ietf.org/html/rfc6578#section-6.1
+type SyncCollectionQuery struct {
+	XMLName   xml.Name `xml:"DAV: sync-collection"`
+	SyncToken string   `xml:"sync-token"`
+	Limit     *Limit   `xml:"limit,omitempty"`
+	SyncLevel string   `xml:"sync-level,omitempty"`
+	Prop      *Prop    `xml:"DAV: prop,omitempty"`
```
emersion, line 415: No need to specify the namespace here since it's the same as `XMLName`'s.

AlmogBaku:
done @emersion  :D

AlmogBaku:
ouch, for some reason the PR got mixed up with #36  which you already merged.... just ignore that

internal/elements.go
```diff
@@ -404,3 +392,18 @@ type Set struct {
 	XMLName xml.Name `xml:"DAV: set"`
 	Prop    Prop     `xml:"prop"`
 }
+
+// https://tools.ietf.org/html/rfc6578#section-6.1
+type SyncCollectionQuery struct {
+	XMLName   xml.Name `xml:"DAV: sync-collection"`
+	SyncToken string   `xml:"sync-token"`
+	Limit     *Limit   `xml:"limit,omitempty"`
+	SyncLevel string   `xml:"sync-level,omitempty"`
```
emersion, line 401: It seems like this element is mandatory, so we should leave out `,omitempty`.

internal/elements.go
```diff
@@ -404,3 +392,18 @@ type Set struct {
...
+	Prop      *Prop    `xml:"prop,omitempty"`
```
emersion, line 402: This is mandatory, so we can drop `,omitempty` and remove the pointer (like in `Set` above).

internal/client.go
```diff
@@ -174,3 +174,32 @@ func (c *Client) Options(path string) (classes map[string]bool, methods map[stri
 	methods = parseCommaSeparatedSet(resp.Header["Allow"], true)
 	return classes, methods, nil
 }
+
+type SyncLevel string
```
emersion, line 178: This type is unused.

internal/client.go
```diff
@@ -174,3 +174,32 @@ func (c *Client) Options(path string) (classes map[string]bool, methods map[stri
...
+
+const (
+	SyncLevel1        SyncLevel = "1"
+	SyncLevelInfinite SyncLevel = "infinite"
+)
+
+// SyncCollection perform a `sync-collection` REPORT operation on a resource
+func (c *Client) SyncCollection(path, syncToken string, level Depth, limit *Limit, prop *Prop) (*Multistatus, error) {
+	q := &SyncCollectionQuery{
```
emersion, line 187: No need to take the address here with `&`, because that results in two indirections with the additional `&` in the `NewXMLRequest` call below.

carddav/client.go
```diff
@@ -429,3 +429,24 @@ func (c *Client) PutAddressObject(path string, card vcard.Card) (*AddressObject,
 	}
 	return ao, nil
 }
+
+// SyncCollection do a sync-collection operation on resource(path), it returns an array of modified/deleted/new
+// AddressObject, the next SyncToken and error
+func (c *Client) SyncCollection(path string, query *SyncQuery) ([]AddressObject, string, error) {
```
emersion, line 435: Returning `[]AddressObject` isn't enough here. The server will return the new/updated resources with a 200 status, and the deleted resources with a 404. Each new/updated resource will have new/updated properties with a 200 status, and deleted properties with a 404.

We need to use a data structure capable of representing this. I think as a first version we could ignore deleted props and use something like this:

```go
type SyncResponse {
	Token string
	Updated []AddressObject
	Deleted []string
}
```

What do you think?

AlmogBaku:
done.

carddav/client.go
```diff
@@ -429,3 +429,62 @@ func (c *Client) PutAddressObject(path string, card vcard.Card) (*AddressObject,
...
+
+	ms, err := c.ic.SyncCollection(path, query.SyncToken, internal.DepthOne, limit, propReq)
+	if err != nil {
+		return nil, err
+	}
+
+	ret := &SyncResponse{}
+	for _, resp := range ms.Responses {
+		p, err := resp.Path()
+		if err != nil {
```
emersion, line 454: I think `Path` returns an error that verifies `internal.IsNotFound` when the status is 404.

Would be nice to actually test that this code works :)

AlmogBaku:
fixed

AlmogBaku:
@emersion I've implemented all of your comments... can you please re-review it?

carddav/carddav.go
```diff
@@ -99,3 +99,17 @@ type AddressObject struct {
...
+	DataRequest AddressDataRequest
+	SyncToken   string
+	Limit       int // <= 0 means unlimited
+}
+
+//SyncResponse contains the returned sync-token for next time
+type SyncResponse struct {
+	SyncToken string
+	Updated   []AddressObject
+	Deleted   []AddressObject
```
emersion, line 114: Deleted address objects will only have a href, no actual data. So it would make sense to make this a list of paths instead (`[]string`).

carddav/client.go
```diff
@@ -429,3 +429,52 @@ func (c *Client) PutAddressObject(path string, card vcard.Card) (*AddressObject,
...
+	}
+
+	ms, err := c.ic.SyncCollection(path, query.SyncToken, internal.DepthOne, limit, propReq)
+	if err != nil {
+		return nil, err
+	}
+
+	ret := &SyncResponse{SyncToken: ms.SyncToken}
+	for _, resp := range ms.Responses {
+		p := resp.Hrefs[0].Path
```
emersion, line 452: There is `Response.Path` which takes care of this logic. `internal.IsNotFound` can be used on the returned error to figure out whether the resource has been deleted.

carddav/client.go
```diff
@@ -429,3 +429,52 @@ func (c *Client) PutAddressObject(path string, card vcard.Card) (*AddressObject,
...
+		if p == path || path == fmt.Sprintf("%s/", p) {
```
emersion, line 453: I don't think the `/` workaround is necessary because address objects will be files, not directories.

AlmogBaku: but the same directory present...
how do you suggest to solve it?

carddav/client.go
AlmogBaku: the reason I chose AddressObject is that I thought it make more sense to return a similar and structured interface

AlmogBaku:
@emersion any comment?

carddav/client.go
emersion: What do you mean?

carddav/client.go
emersion: No, properties are never returned for deleted objects.

carddav/client.go
AlmogBaku: without this practice, I still get the path concatenated with a trailing slash.

emersion: Hmm. Alright, I don't see a better solution, let's keep the `/` workaround then.

AlmogBaku:
done. @emersion 

Let's merge 🎉

carddav/client.go
emersion: Opened https://github.com/emersion/go-webdav/issues/41

PR #42, opened 2020-09-06
Improve service discovery, implement SD for CalDav

### LINKED ISSUE #30: carddav: add support for well-known URI
See https://tools.ietf.org/html/rfc6764#section-4

Related: https://github.com/emersion/hydroxide/issues/44

dpeukert:
Closes #30 

This is my first time writing Go code, so I apologise in advance for any beginner mistakes I made in this :smiley:.


internal/client.go
```diff
@@ -6,13 +6,98 @@ import (
...
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return "", err
+	}
+	resp.Body.Close()
+
+	// Servers might require authentication to perform an OPTIONS request
+	if resp.StatusCode/100 != 2 && resp.StatusCode != http.StatusUnauthorized {
+		return "", fmt.Errorf("HTTP request to %v failed: %v %v", serviceUrl, resp.StatusCode, resp.Status)
+	}
```
dpeukert, line 96: I borrowed this from alps

dpeukert:
Build failure seems unrelated.

internal/client.go
```diff
@@ -6,13 +6,98 @@ import (
...
+	// If we didn't get a path from TXT records, use the default well-known location
+	if path == "" {
+		path = fmt.Sprintf("/.well-known/%v", service)
+	}
+
+	u := url.URL{Scheme: "https", Host: host, Path: path}
+	serviceUrl := u.String()
+
+	// Check if the resulting URL hosts a service
+	req, err := http.NewRequest(http.MethodOptions, serviceUrl, nil)
```
emersion, line 82: [RFC 6764 section 6](https://tools.ietf.org/html/rfc6764#section-6) mandates that a PROPFIND request should be made with the username/password set.

Alps has a different requirement: it needs to sanity check the URLs on startup (without any username/password).

To follow the RFC, we could:

- Take an `webdav.HTTPClient` to handle auth
- Change OPTIONS to a PROPFIND request with a `DAV:current-user-principal` prop
- Return the `DAV:current-user-principal` prop

emersion: To accomodate for Alps, we could have an intermediate `DiscoverContextURL` function that follows the RFC up to and not including the PROPFIND request.

carddav/client.go
```diff
@@ -16,36 +15,9 @@ import (
...
-
-	u := url.URL{Scheme: "https"}
-	if addr.Port == 443 {
-		u.Host = target
-	} else {
-		u.Host = fmt.Sprintf("%v:%v", target, addr.Port)
-	}
-	return u.String(), nil
+func Discover(host string) (string, error) {
+	url, err := internal.Discover("carddav", host)
```
emersion, line 19: Nit: you can just `return internal.Discover(…)` here

emersion (changes requested):
Thanks for working on this! I left a few comments.

carddav/client.go
```diff
@@ -16,36 +15,9 @@ import (
 	"github.com/emersion/go-webdav/internal"
 )
 
-// Discover performs a DNS-based CardDAV service discovery as described in
-// RFC 6352 section 11. It returns the URL to the CardDAV server.
```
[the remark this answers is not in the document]
dpeukert: Fixed in 96034fb

dpeukert:
Sorry about the delay on this, I'll try to get the remaining change sorted this weekend.

emersion:
7e076258d6c10ef3227f38399423c09e7f2555db adds discovery for CalDAV. What's still missing is the TXT DNS record stuff.

PR #43, opened 2020-09-09
fix: deprecrated conversion from int64 to string

proletarius101:
Same as emersion/go-smtp#117


internal/client.go
```diff
@@ -195,7 +195,7 @@ func (c *Client) Options(path string) (classes map[string]bool, methods map[stri
 func (c *Client) SyncCollection(path, syncToken string, level Depth, limit *Limit, prop *Prop) (*Multistatus, error) {
 	q := SyncCollectionQuery{
 		SyncToken: syncToken,
-		SyncLevel: string(level),
+		SyncLevel: string(rune(level)),
```
emersion, line 198: In this case we don't want to convert to an ASCII character, we want to format the level as a number. `level.String()` is designed for this.

proletarius101: Oh, that's fixed now

PR #44, opened 2020-09-23
return a nil CalendarCompRequest when empty

kwo:
Hi. I'm trying to use QueryCalendar with iCloud calendars but am not to do it without these changes. If a CalendarCompRequest if not included in a calendar query, it will be initialized as a blank/empty struct. The code now tests if the CalendarCompRequest is empty and if so returns a nil `comp` so that the `calendarDataReq` can be sent properly without a Comp section.


caldav/caldav.go
```diff
@@ -58,3 +58,8 @@ type CalendarObject struct {
 	ETag    string
 	Data    *ical.Calendar
 }
+
+// IsEmpty tests if a CalendarCompRequest contains any information
+func (z CalendarCompRequest) IsEmpty() bool {
```
emersion, line 63: Maybe we can just check `z.Name == ""` and maybe we don't need to introduce `IsEmpty`? My reading of [the RFC](https://tools.ietf.org/html/rfc4791#section-9.6.1) suggests it's actually the only required field.

emersion:
The idea looks good to me. Maybe we could instead change the `CalendarCompRequest` fields to be pointers, but not sure it's worth it.

emersion:
Hmm. Actually, why is `CalendarCompRequest.Name` not set to `VCALENDAR`? [The RFC](https://tools.ietf.org/html/rfc4791#section-9.7.1) seems to indicate that `<filter>` MUST have a `<comp-filter>`, and that `<comp-filter>` MUST have a `name` attribute.

caldav/caldav.go
kwo: remove IsEmpty

kwo:
> The idea looks good to me. Maybe we could instead change the `CalendarCompRequest` fields to be pointers, but not sure it's worth it.

That would also work

emersion:
>Hmm. Actually, why is CalendarCompRequest.Name not set to VCALENDAR? The RFC seems to indicate that <filter> MUST have a <comp-filter>, and that <comp-filter> MUST have a name attribute.

Thoughts?

PR #45, opened 2020-10-08
Add handling for carddav/caldav well-known urls

miterion:
This pr adds handling of well-known URLs for both carddav and caldav. According to the [RFC 6764, Section 6](https://tools.ietf.org/html/rfc6764#section-6) the initial request for a client is a PROPFIND against `well-known/carddav`or `well-known/caldav`, which the server should reply to with a redirect to the proper endpoint (which here corresponds to the "/" url).

This might partially solve #30

### DIFF
--- a/carddav/server.go
+++ b/carddav/server.go
@@ -36,6 +36,11 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if r.URL.Path == "/.well-known/carddav" {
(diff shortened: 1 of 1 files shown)


emersion (commented):
Thanks for sending a PR!

Can we handle this in `carddav/server.go` in `Handler.ServeHTTP` instead? This would avoid hard-coding CardDAV/CalDAV support in the common code.

miterion:
In this case I think it makes sense to limit the check to the well known URL for carddav. Will fix that tomorrow

emersion (approved):
Thanks!

PR #46, opened 2021-01-09
Return href together with a 404

Apehaenger:
In case of a card deletion, SyncCollection() responded with a SyncResponse.Deleted: [ "" ].
Instead of, a href should be returned together with a 404. Please see https://tools.ietf.org/html/rfc6578#section-3.2

The reason was the early exit (without a href/path) of Response.Path() in case of an error/404.

I slightly restructured Response.path() so that it remains simple and readable (without cascaded if's)

### DIFF
--- a/internal/elements.go
+++ b/internal/elements.go
@@ -134,17 +134,18 @@ func NewOKResponse(path string) *Response {
 	}
 }
 
-func (resp *Response) Path() (path string, err error) {
-	if len(resp.Hrefs) == 1 {
-		path = resp.Hrefs[0].Path // A sync response for a deleted entry has to have a href/path together with a 404. See https://tools.ietf.org/html/rfc6578#section-3.2
-	}
-	if err = resp.Status.Err(); err != nil {
-		return
+func (resp *Response) Path() (string, error) {
+	if err := resp.Status.Err(); err != nil {
+		if len(resp.Hrefs) == 1 {
+			path := resp.Hrefs[0].Path // A sync response for a deleted entry has to have a path together with a 404. See https://tools.ietf.org/html/rfc6578#section-3.2
+			return path, err
+		}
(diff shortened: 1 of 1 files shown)


internal/elements.go
```diff
@@ -134,14 +134,17 @@ func NewOKResponse(path string) *Response {
 	}
 }
 
-func (resp *Response) Path() (string, error) {
-	if err := resp.Status.Err(); err != nil {
-		return "", err
+func (resp *Response) Path() (path string, err error) {
```
emersion, line 137: Please don't use naked returns. Also no need for named result parameters here, it's clear what these are from the function name alone.

See https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters

emersion (changes requested):
Looks good to me apart from this style issue.

internal/elements.go
Apehaenger: No problem.
I felt it look ugly and quirky with the named returns due to the cascaded if's... or I'm not smart enough ;-)

emersion: No named return doesn't mean you need to give up the duplicated if:

```go
err := resp.Status.Err()
var path string
if len(resp.Hrefs) == 1 {
	path = resp.Hrefs[0].Path
} else if err == nil {
	err = fmt.Errorf("webdav: malformed response: expected exactly one href element, got %v", len(resp.Hrefs))
}
return path, err
```

Apehaenger: Quite cool!
I definitely wasn't smart enough ;-)
My brain wasn't able to find the "if len ... else err ..." variant :-/

Thanks a lot for the simplicity lesson and sorry for the effort for such a simple pull request!

emersion (approved):
Thanks for your contribution!

PR #49, opened 2021-03-18
carddav: add filtering and matching helper functions

sbinet:
Updates emersion/hydroxide#159.

Signed-off-by: Sebastien Binet <binet@cern.ch>

### DIFF
--- a/carddav/match.go
+++ b/carddav/match.go
@@ -1,0 +1,149 @@
+package carddav
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/emersion/go-vcard"
+)
+
+// Filter returns the filtered list of address objects matching the provided query.
+// A nil query will return the full list of address objects.
+func Filter(query *AddressBookQuery, aos []AddressObject) ([]AddressObject, error) {
+	if query == nil {
+		// FIXME: should we always return a copy of the provided slice?
+		return aos, nil
+	}
+
+	n := query.Limit
+	if n <= 0 {
+		n = len(aos)
+	}
+	out := make([]AddressObject, 0, n)
+	for _, ao := range aos {
+		ok, err := Match(query, ao)
+		if err != nil {
+			return nil, err
+		}
+		if !ok {
+			continue
+		}
+		out = append(out, ao)
+		if len(out) >= n {
+			break
+		}
+	}
+	return out, nil
+}
+
+// Match reports whether the provided AddressObject matches the query.
+func Match(query *AddressBookQuery, ao AddressObject) (matched bool, err error) {
+	if query == nil {
+		return true, nil
+	}
+
+	if query.DataRequest.AllProp {
+		for _, name := range query.DataRequest.Props {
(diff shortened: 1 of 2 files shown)


sbinet:
gentle ping :)

carddav/match.go
```diff
@@ -0,0 +1,149 @@
...
+		out = append(out, ao)
+		if len(out) >= n {
+			break
+		}
+	}
+	return out, nil
+}
+
+// Match reports whether the provided AddressObject matches the query.
+func Match(query *AddressBookQuery, ao AddressObject) (matched bool, err error) {
```
emersion, line 40: Maybe should pass `AddressObject` as a pointer here?

carddav/match.go
```diff
@@ -0,0 +1,149 @@
...
+	if query == nil {
+		return true, nil
+	}
+
+	if query.DataRequest.AllProp {
```
emersion, line 45: `AllProp` means that the client wants all properties to be returned, ie. `Props` is ignored. Since we don't filter returned properties yet (we always return all props), we can ignore it.

This can probably be turned into a TODO comment.

carddav/match.go
```diff
@@ -0,0 +1,149 @@
...
+			if !ok {
+				return false, nil
+			}
+		}
+		return true, nil
+	}
+}
+
+func matchPropFilter(prop PropFilter, ao AddressObject) (bool, error) {
+	field := ao.Card.Get(prop.Name)
```
emersion, line 86: This only matches the filter against the first property. I think the correct behaviour would be to match against all fields which have the name, can we add that as a TODO?

carddav/match.go
```diff
@@ -0,0 +1,149 @@
...
+	if field == nil {
+		// assume AddressBookQuery.DataRequest.AllProp is false
```
emersion, line 88: See my comment above about `AllProp`.

I think this branch is still doing the right thing. The comment is a little confusing.

carddav/match.go
```diff
@@ -0,0 +1,149 @@
...
+			if !ok {
+				return false, nil
+			}
+		}
+		return true, nil
+	}
+}
+
+func matchTextMatch(txt TextMatch, field *vcard.Field) (bool, error) {
+	// TODO: handle carddav.TextMatch.IsNotDefined.
```
emersion, line 126: Hm, does `TextMatch.IsNotDefined` exist? This sounds weird looking at the RFC.

emersion, line 126: Can we add a TODO about collation handling? By default, we should compare with "i;unicode-casemap", but the client can also specify other collations (such as a stricter bytewise comparison like we do here).

emersion (changes requested):
Thanks for the pull request! Here are some comments.

I'm completely fine with leaving some things unimplemented or not 100% correct at first, but I'd like to add some TODOs so that we don't forget about these.

emersion:
Gentle ping

emersion:
Superseded by #55

PR #51, opened 2021-08-11
Include ModTime for directories if available

jumo98:
Hi!
I'm currently working on a WebDAV project and noticed that the `modTime` is not included for directories.
The code now includes the `modTime`, if available in the XML.

### DIFF
--- a/client.go
+++ b/client.go
@@ -92,8 +92,11 @@ func fileInfoFromResponse(resp *internal.Response) (*FileInfo, error) {
 			return nil, err
 		}
 
+		if !internal.IsNotFound(err) {
+			fi.ModTime = time.Time(getMod.LastModified)
+		}
+
 		fi.IsDir = true
-		fi.ModTime = time.Time(getMod.LastModified)
 	} else {
 		var getLen internal.GetContentLength
 		if err := resp.DecodeProp(&getLen); err != nil {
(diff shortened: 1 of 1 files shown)


client.go
```diff
@@ -87,6 +87,15 @@ func fileInfoFromResponse(resp *internal.Response) (*FileInfo, error) {
 		return nil, err
 	}
 	if resType.Is(internal.CollectionName) {
+		var getMod internal.GetLastModified
+		if err := resp.DecodeProp(&getMod); err != nil && !internal.IsNotFound(err) {
+			return nil, err
+		}
+
+		if !internal.IsNotFound(err) {
```
emersion, line 95: `err` here is not the same variable as the one if the `if` above, because of scoping.

emersion (commented):
Looks good! One last thing: since there's now the same code for both directories and regular files, maybe we can move it outside of the `if resType.Is(internal.CollectionName)`?

jumo98:
Good idea!
Moved it outside.

emersion:
There's one remaining `getMod` variable in the `else` branch, which is now unused. Can we get rid of it?

jumo98:
Done.

emersion (approved):
LGTM, thanks!

PR #53, opened 2022-02-23
Pass request context to backend interface

bitfehler:
This way the storage implementation can communicate with any potentially
used middleware (e.g. authentication) or for example abort requests.


emersion (approved):
Thanks!

PR #54, opened 2022-02-24
Set ETag and Last-Modified if available

bitfehler:
Some clients (e.g. Evolution) will not work properly without this. It is
up to the underlying backend to actually provide this data, the headers
will only be set if it is available.

### DIFF
--- a/carddav/server.go
+++ b/carddav/server.go
@@ -262,7 +262,12 @@ func (b *backend) HeadGet(w http.ResponseWriter, r *http.Request) error {
 	}
 
 	w.Header().Set("Content-Type", vcard.MIMEType)
-	// TODO: set ETag, Last-Modified
+	if ao.ETag != "" {
+		w.Header().Set("ETag", internal.ETag(ao.ETag).String())
+	}
+	if !ao.ModTime.IsZero() {
+		w.Header().Set("Last-Modified", ao.ModTime.UTC().Format(http.TimeFormat))
+	}
 
 	if r.Method != http.MethodHead {
 		return vcard.NewEncoder(w).Encode(ao.Card)


carddav/server.go
```diff
@@ -262,7 +262,12 @@ func (b *backend) HeadGet(w http.ResponseWriter, r *http.Request) error {
 	}
 
 	w.Header().Set("Content-Type", vcard.MIMEType)
-	// TODO: set ETag, Last-Modified
+	if ao.ETag != "" {
+		w.Header().Set("ETag", ao.ETag)
+	}
+	if !ao.ModTime.IsZero() {
+		w.Header().Set("Last-Modified", ao.ModTime.Format(http.TimeFormat))
```
emersion, line 269: From the `net/http` docs:

> The time being formatted must be in UTC for Format to generate the correct format. 

So need to add an `.UTC()` in there, in case the backend returns a non-UTC date.

carddav/server.go
```diff
@@ -262,7 +262,12 @@ func (b *backend) HeadGet(w http.ResponseWriter, r *http.Request) error {
 	}
 
 	w.Header().Set("Content-Type", vcard.MIMEType)
-	// TODO: set ETag, Last-Modified
+	if ao.ETag != "" {
+		w.Header().Set("ETag", ao.ETag)
```
emersion, line 266: The ETag needs to be quoted here. The marshalling can be done via: `internal.ETag(ao.ETag).String()`.

See `backend.HeadGet` in `/server.go` for a reference.

emersion (approved):
LGTM, thanks!

Next time, please add a "carddav: " prefix to the commit message, since this only touches carddav stuff. :)

bitfehler:
Will do! Thanks a lot!

PR #55, opened 2022-02-28
carddav: add filtering and matching helper functions

bitfehler:
This is a continuation of #49

I addressed all your comments and removed two test cases that I think were a misunderstanding about the semantics of `DataRequest.AllProp`.

I did add handling of `carddav.PropFilter.IsNotDefined` because it seemed simple, let me know if I got that wrong there.

About the collation handling, I added the TODO, but it seems this needs support in the `TextMatch` type first, correct?

### DIFF
--- a/carddav/match.go
+++ b/carddav/match.go
@@ -1,0 +1,146 @@
+package carddav
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/emersion/go-vcard"
+)
+
+// Filter returns the filtered list of address objects matching the provided query.
+// A nil query will return the full list of address objects.
+func Filter(query *AddressBookQuery, aos []AddressObject) ([]AddressObject, error) {
+	if query == nil {
+		// FIXME: should we always return a copy of the provided slice?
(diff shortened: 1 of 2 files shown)


carddav/match.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+}
+
+func matchPropFilter(prop PropFilter, ao *AddressObject) (bool, error) {
+	// TODO: this only matches first field, there could be multiple
+	field := ao.Card.Get(prop.Name)
+	if field == nil {
+		return prop.IsNotDefined, nil
+	}
+
```
emersion, line 84: At this point `field != nil` so we maybe need to return false in case `prop.IsNotDefined` is set?

emersion:
Apart from this comment, looks pretty good to me!

carddav/match.go
bitfehler: Good catch, thanks! Fixed.

emersion (approved):
LGTM, thanks!

PR #56, opened 2022-03-01
Add error type representing DAV/XML errors

bitfehler:
Backends will need some way to signal that a precondition error occurred
(and specifying which one) without causing the server to return a 500.
This commit adds an exported function to create a specific error for
this. The existing error handling routine is slightly adapted to handle
this error in such a way that it (almost) returns the desired result.

Usage would be something like:

    return "", carddav.PreconditionError("no-uid-conflict")

As is, this approach is missing a few things to be correct:

* Content-Type response header is text/plain, should be text/xml
* Response body is missing XML header (`<?xml ...`)

I'd like to get your opinion as early as possible before I start writing
more code.

Other approaches I could think of would be creating another, more
specific error type for this, using the context for communicating
specifics, or adding a third return value to the relevant interface
funcions.

What I think is nice about this approach is that it leaves the
interfaces as is and does not require handling of this kind of errors in
all the places they could occur.

What do you think?

### DIFF
--- a/carddav/server.go
+++ b/carddav/server.go
@@ -439,3 +439,28 @@ func (b *backend) Copy(r *http.Request, dest *internal.Href, recursive, overwrit
 func (b *backend) Move(r *http.Request, dest *internal.Href, overwrite bool) (created bool, err error) {
 	panic("TODO")
 }
+
+// https://tools.ietf.org/rfcmarkup?doc=6352#section-6.3.2.1
+type PreconditionType string
+
+const (
+	PreconditionNoUIDConflict        PreconditionType = "no-uid-conflict"
+	PreconditionSupportedAddressData PreconditionType = "supported-address-data"
+	PreconditionValidAddressData     PreconditionType = "valid-address-data"
+	PreconditionMaxResourceSize      PreconditionType = "max-resource-size"
+)
+
+func NewPreconditionError(err PreconditionType) error {
+	name := xml.Name{"urn:ietf:params:xml:ns:carddav", string(err)}
+	elem := internal.NewRawXMLElement(name, nil, nil)
+	e := internal.Error{
+		Raw: []internal.RawXMLValue{
+			*elem,
+		},
+	}
+	return &internal.DAVError{
+		Code: 409,
+		Msg:  fmt.Sprintf("precondition not met: %s", string(err)),
+		Err:  e,
+	}
+}
--- a/internal/server.go
+++ b/internal/server.go
@@ -10,6 +10,12 @@ import (
 )
 
 func ServeError(w http.ResponseWriter, err error) {
+	if davErr, ok := err.(*DAVError); ok {
+		w.WriteHeader(davErr.Code)
+		ServeXML(w).Encode(davErr.Err)
+		return
+	}
+
 	code := http.StatusInternalServerError
 	if httpErr, ok := err.(*HTTPError); ok {
 		code = httpErr.Code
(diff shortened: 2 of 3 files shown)


bitfehler:
Of course this would not just be for backends. I think it would help take care of all these TODOs: https://github.com/emersion/go-webdav/blob/master/carddav/server.go#L406-L421

bitfehler:
Totally agree, did this now. Also moved some stuff around, let me know if you'd like to see something elsewhere. I also update the commit message with some details. In general I think this will be a useful mechanism. Of course something might show up down the road that might require some modification, but I guess that's life...

internal/server.go
```diff
@@ -3,13 +3,21 @@ package internal
...
 	"net/http"
 	"net/url"
 	"strings"
 )
 
 func ServeError(w http.ResponseWriter, err error) {
+	if davErr, ok := err.(*DAVError); ok {
+		w.Header().Set("Content-Type", "text/xml; charset=utf-8")
+		w.WriteHeader(davErr.Code)
+		io.WriteString(w, davErr.Error())
```
emersion, line 17: I think this is missing the XML header (`<?xml version="1.0" encoding="UTF-8"?>`).

We can just use `ServeXML` instead, which takes care of that + setting the Content-Type.

emersion: Ah, the header is added in `Error`. Still, I'd prefer to use `ServeXML` here.

carddav/server.go
```diff
@@ -439,3 +439,19 @@ func (b *backend) Copy(r *http.Request, dest *internal.Href, recursive, overwrit
 func (b *backend) Move(r *http.Request, dest *internal.Href, overwrite bool) (created bool, err error) {
 	panic("TODO")
 }
+
+func PreconditionError(err string) error {
```
emersion, line 443: We should probably create a separate type for precondition error types, e.g.

```go
type PreconditionErrorType string

const (
    PreconditionErrorNoUIDConflict PreconditionErrorType = "no-uid-conflict"
)
```

to avoid potential typos in backends

carddav/server.go
```diff
@@ -439,3 +439,19 @@ func (b *backend) Copy(r *http.Request, dest *internal.Href, recursive, overwrit
...
+	elem := internal.NewRawXMLElement(
+		xml.Name{"urn:ietf:params:xml:ns:carddav", err},
+		[]xml.Attr{}, []internal.RawXMLValue{})
```
emersion, line 446: `nil` can be used instead of explicit empty slices. This is shorter and avoids allocations.

carddav/server.go
```diff
@@ -439,3 +439,29 @@ func (b *backend) Copy(r *http.Request, dest *internal.Href, recursive, overwrit
...
+type PreconditionType string
+
+const (
+	PreconditionNoUIDConflict        PreconditionType = "no-uid-conflict"
+	PreconditionSupportedAddressData PreconditionType = "supported-address-data"
+	PreconditionValidAddressData     PreconditionType = "valid-address-data"
+	PreconditionMaxResourceSize      PreconditionType = "max-resource-size"
+)
+
+func PreconditionError(err PreconditionType) error {
```
emersion, line 453: Ah, I missed this in my first review: can we name this `NewPreconditionError` instead? This will allow us to introduce a `PreconditionError` type without conflicts in the future if we need to.

carddav/server.go
```diff
@@ -439,3 +439,29 @@ func (b *backend) Copy(r *http.Request, dest *internal.Href, recursive, overwrit
...
+	name := xml.Name{"urn:ietf:params:xml:ns:carddav", string(err)}
+	elem := internal.NewRawXMLElement(name, nil, nil)
+	e := internal.Error{
+		XMLName: xml.Name{"DAV", "error"},
```
emersion, line 457: Nit: I don't think we need to specify this field, Go should use these values by default.

bitfehler: I think I don't fully understand that, but it works indeed... ¯\\\_(ツ)_/¯

emersion (approved):
LGTM, thanks!

PR #57, opened 2022-03-11
carddav: allow setting base endpoint other than /

bitfehler:
This way the handler can work on e.g. `/carddav/`. The .well-known
handler can even be re-used. With e.g. chi, this works:

```
mux := chi.NewRouter()
handler := carddav.Handler{Backend: backend, Prefix: "/carddav/"}
mux.Method("PROPFIND", "/.well-known/carddav", &handler)
mux.Mount("/carddav/", &handler)
```

Some care has to be taken. For example, using `"/carddav"` (no trailing
slash) will not work. Also, any backend implementation must be able to
handle the additional prefix.

### DIFF
--- a/carddav/server.go
+++ b/carddav/server.go
@@ -28,6 +28,7 @@ type Backend interface {
 // server.
 type Handler struct {
 	Backend Backend
+	Prefix  string
 }
 
 // ServeHTTP implements http.Handler.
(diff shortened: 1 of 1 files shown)


bitfehler:
One minute too late, I realized: should we wrap the handler creation in a function that also asserts that the prefix starts and ends with a slash?

bitfehler:
A better solution was found after discussion

PR #58, opened 2022-03-16
Bitfehler/caldav server

bitfehler:
These are essentially your commits from the `caldav-report` and `caldav-server` branches. I only reworded the commit messages and removed the debug output (for now, I have that on another branch and would bring it back with a flag or such).

The result does offer a very basic server. Most operations are not implemented yet, but those that are do work.

If you are okay to merge this I would have a few more changes on top of this.

### DIFF
--- a/caldav/elements.go
+++ b/caldav/elements.go
@@ -2,6 +2,7 @@ package caldav
 
 import (
 	"encoding/xml"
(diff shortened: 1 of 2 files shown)


emersion (approved):
Thanks for cleaning this up!

PR #59, opened 2022-03-16
Pass request context to backend interface

bitfehler:
This aligns the caldav interface with the carddav one (see #53).


emersion (approved):
Thanks!

PR #60, opened 2022-03-17
carddav: pass If-(None-)Match to backend

bitfehler:
This simply extends the interfaces to pass on the values if they were
used, relying on the backend to handle things accordingly.

### DIFF
--- a/carddav/server.go
+++ b/carddav/server.go
@@ -14,13 +14,22 @@ import (
 
 // TODO: add support for multiple address books
 
+type PutAddressObjectOptions struct {
+	// IfNoneMatch indicates that the client does not want to overwrite
+	// an existing resource.
+	IfNoneMatch bool
+	// IfMatch provides the ETag of the resource that the client intends
+	// to overwrite, can be ""
+	IfMatch string
+}
+
 // Backend is a CardDAV server backend.
 type Backend interface {
 	AddressBook(ctx context.Context) (*AddressBook, error)
 	GetAddressObject(ctx context.Context, path string, req *AddressDataRequest) (*AddressObject, error)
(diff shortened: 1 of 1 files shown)


carddav/server.go
```diff
@@ -397,7 +397,8 @@ func (b *backend) Proppatch(r *http.Request, update *internal.Propertyupdate) (*
 }
 
 func (b *backend) Put(r *http.Request) (*internal.Href, error) {
-	// TODO: add support for If-None-Match and If-Match
+	ifNoneMatch := r.Header.Get("If-None-Match") == "*"
```
emersion, line 400: Can we error out if `If-None-Match` is set, but not `*`?

carddav/server.go
```diff
@@ -20,7 +20,7 @@ type Backend interface {
 	GetAddressObject(ctx context.Context, path string, req *AddressDataRequest) (*AddressObject, error)
 	ListAddressObjects(ctx context.Context, req *AddressDataRequest) ([]AddressObject, error)
 	QueryAddressObjects(ctx context.Context, query *AddressBookQuery) ([]AddressObject, error)
-	PutAddressObject(ctx context.Context, path string, card vcard.Card) (loc string, err error)
+	PutAddressObject(ctx context.Context, path string, card vcard.Card, ifNoneMatch bool, ifMatch string) (loc string, err error)
```
emersion, line 23: I wonder if we should introduce a `PutAddressObjectOptions` struct

emersion (approved):
LGTM! Pushed a very minor edit to pass the struct as a pointer, since the struct may grow in the future and this is what the stdlib does.

Thanks!

Issue #27, opened 2020-01-25
Aggregate multiple servers


emersion:
It may be desirable to have a base WebDAV filesystem, with some CalDAV/CardDAV endpoints. This should work fine except for [principals](https://tools.ietf.org/html/rfc5397). We need to figure out a way to share principals between servers.

PR #62, opened 2022-03-23
Support custom user principal and home set paths

bitfehler:
Currently, the user principal path and the home set path are both
hardcoded to "/", for both CalDAV and CardDAV. This poses a challenge if
one wishes to run a CardDAV and CalDAV server in the same server.

This commit introduces the concept of a UserPrincipalBackend. This
backend must provide the locations for the user principal as well as the
home sets. The CardDAV and CalDAV servers act accordingly.

The individual servers will continue to work as before (including the
option of keeping everything at "/"). If one wishes to run CardDAV and
CalDAV in parallel, the new `webdav.ServeUserPrincipal()` can be used,
which will use the UserPrincipalBackend interface to serve a common user
principal for both servers (if so desired, can be used for a single one
also).

Note that the storage backend will have to know about these paths as
well. For any non-trivial use case, a storage backend should probably
have access to the same UserPrincipalBackend. That is, however, an
implementation detail and doesn't have to be reflected in the
interfaces.

### DIFF
--- a/caldav/server.go
+++ b/caldav/server.go
@@ -8,7 +8,7 @@ import (
 	"time"
 
 	"github.com/emersion/go-ical"
-
+	"github.com/emersion/go-webdav"
 	"github.com/emersion/go-webdav/internal"
 )
 
@@ -16,10 +16,13 @@ import (
 
 // Backend is a CalDAV server backend.
 type Backend interface {
+	CalendarHomeSetPath(ctx context.Context) (string, error)
 	Calendar(ctx context.Context) (*Calendar, error)
 	GetCalendarObject(ctx context.Context, path string, req *CalendarCompRequest) (*CalendarObject, error)
 	ListCalendarObjects(ctx context.Context, req *CalendarCompRequest) ([]CalendarObject, error)
 	QueryCalendarObjects(ctx context.Context, query *CalendarQuery) ([]CalendarObject, error)
+
+	webdav.UserPrincipalBackend
 }
 
 // Handler handles CalDAV HTTP requests. It can be used to create a CalDAV
@@ -35,12 +38,17 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if r.URL.Path == "/.well-known/caldav" {
-		http.Redirect(w, r, "/", http.StatusMovedPermanently)
+	principalPath, err := h.Backend.CurrentUserPrincipal(r.Context())
+	if err != nil {
+		http.Error(w, "caldav: failed to determine current user principal", http.StatusInternalServerError)
 		return
 	}
 
-	var err error
+	if r.URL.Path == "/.well-known/caldav" {
+		http.Redirect(w, r, principalPath, http.StatusMovedPermanently)
+		return
+	}
+
 	switch r.Method {
 	case "REPORT":
 		err = h.handleReport(w, r)
@@ -182,7 +190,17 @@ type backend struct {
 func (b *backend) Options(r *http.Request) (caps []string, allow []string, err error) {
 	caps = []string{"calendar-access"}
 
-	if r.URL.Path == "/" {
+	homeSetPath, err := b.Backend.CalendarHomeSetPath(r.Context())
+	if err != nil {
+		return nil, nil, err
+	}
+
+	principalPath, err := b.Backend.CurrentUserPrincipal(r.Context())
+	if err != nil {
+		return nil, nil, err
+	}
+
+	if r.URL.Path == "/" || r.URL.Path == principalPath || r.URL.Path == homeSetPath {
 		return caps, []string{http.MethodOptions, "PROPFIND", "REPORT"}, nil
 	}
 
@@ -209,14 +227,30 @@ func (b *backend) HeadGet(w http.ResponseWriter, r *http.Request) error {
 }
 
 func (b *backend) Propfind(r *http.Request, propfind *internal.Propfind, depth internal.Depth) (*internal.Multistatus, error) {
+	homeSetPath, err := b.Backend.CalendarHomeSetPath(r.Context())
+	if err != nil {
+		return nil, err
+	}
+	principalPath, err := b.Backend.CurrentUserPrincipal(r.Context())
+	if err != nil {
+		return nil, err
+	}
+
 	var resps []internal.Response
-	if r.URL.Path == "/" {
+
+	if r.URL.Path == principalPath {
+		resp, err := b.propfindUserPrincipal(r.Context(), propfind, homeSetPath)
+		if err != nil {
+			return nil, err
+		}
+		resps = append(resps, *resp)
+	} else if r.URL.Path == homeSetPath {
 		cal, err := b.Backend.Calendar(r.Context())
 		if err != nil {
 			return nil, err
 		}
 
-		resp, err := b.propfindCalendar(propfind, cal)
+		resp, err := b.propfindCalendar(r.Context(), propfind, cal)
 		if err != nil {
 			return nil, err
 		}
@@ -225,13 +259,38 @@ func (b *backend) Propfind(r *http.Request, propfind *internal.Propfind, depth i
 		if depth != internal.DepthZero {
 			// TODO
 		}
+	} else {
+		// TODO
 	}
 
 	return internal.NewMultistatus(resps...), nil
 }
 
-func (b *backend) propfindCalendar(propfind *internal.Propfind, cal *Calendar) (*internal.Response, error) {
+func (b *backend) propfindUserPrincipal(ctx context.Context, propfind *internal.Propfind, homeSetPath string) (*internal.Response, error) {
+	principalPath, err := b.Backend.CurrentUserPrincipal(ctx)
+	if err != nil {
+		return nil, err
+	}
 	props := map[xml.Name]internal.PropfindFunc{
+		internal.CurrentUserPrincipalName: func(*internal.RawXMLValue) (interface{}, error) {
+			return &internal.CurrentUserPrincipal{Href: internal.Href{Path: principalPath}}, nil
+		},
+		calendarHomeSetName: func(*internal.RawXMLValue) (interface{}, error) {
+			return &calendarHomeSet{Href: internal.Href{Path: homeSetPath}}, nil
+		},
+	}
+	return internal.NewPropfindResponse(principalPath, propfind, props)
+}
+
+func (b *backend) propfindCalendar(ctx context.Context, propfind *internal.Propfind, cal *Calendar) (*internal.Response, error) {
+	principalPath, err := b.Backend.CurrentUserPrincipal(ctx)
+	if err != nil {
+		return nil, err
+	}
+	props := map[xml.Name]internal.PropfindFunc{
+		internal.CurrentUserPrincipalName: func(*internal.RawXMLValue) (interface{}, error) {
+			return &internal.CurrentUserPrincipal{Href: internal.Href{Path: principalPath}}, nil
+		},
 		internal.ResourceTypeName: func(*internal.RawXMLValue) (interface{}, error) {
 			return internal.NewResourceType(internal.CollectionName, calendarName), nil
 		},
@@ -251,14 +310,6 @@ func (b *backend) propfindCalendar(propfind *internal.Propfind, cal *Calendar) (
 					{Name: ical.CompEvent},
 				},
 			}, nil
-		},
-		// TODO: this is a principal property
-		calendarHomeSetName: func(*internal.RawXMLValue) (interface{}, error) {
-			return &calendarHomeSet{Href: internal.Href{Path: "/"}}, nil
-		},
-		// TODO: this should be set on all resources
-		internal.CurrentUserPrincipalName: func(*internal.RawXMLValue) (interface{}, error) {
-			return &internal.CurrentUserPrincipal{Href: internal.Href{Path: "/"}}, nil
 		},
 	}
 
--- a/carddav/server.go
+++ b/carddav/server.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 
 	"github.com/emersion/go-vcard"
+	"github.com/emersion/go-webdav"
 	"github.com/emersion/go-webdav/internal"
 )
 
@@ -25,12 +26,15 @@ type PutAddressObjectOptions struct {
 
 // Backend is a CardDAV server backend.
 type Backend interface {
+	AddressbookHomeSetPath(ctx context.Context) (string, error)
 	AddressBook(ctx context.Context) (*AddressBook, error)
 	GetAddressObject(ctx context.Context, path string, req *AddressDataRequest) (*AddressObject, error)
 	ListAddressObjects(ctx context.Context, req *AddressDataRequest) ([]AddressObject, error)
 	QueryAddressObjects(ctx context.Context, query *AddressBookQuery) ([]AddressObject, error)
 	PutAddressObject(ctx context.Context, path string, card vcard.Card, opts *PutAddressObjectOptions) (loc string, err error)
 	DeleteAddressObject(ctx context.Context, path string) error
+
+	webdav.UserPrincipalBackend
 }
 
 // Handler handles CardDAV HTTP requests. It can be used to create a CardDAV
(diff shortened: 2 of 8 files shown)


bitfehler:
With this, I have successfully created a single server on which Evolution finds both a calendar as well as an address book. I also tested that using e.g. a simple `carddav.Handler` by itself still works with a UserPrincipalBackend implementation that always returns `/`. Forgot to mention in commit message, but this should take care of #27.

I also managed to create a setup with dynamic user principal paths, so you could have e.g. `/emersion/`, I would have `/bitfehler/`. The storage backend must be able to handle this, of course.

server.go
```diff
@@ -234,3 +237,131 @@ func (b *backend) Move(r *http.Request, dest *internal.Href, overwrite bool) (cr
...
+	relPath, err := RelativeResourcePath(r, backend)
+	if err != nil {
+		return false, err
+	}
+	return relPath != r.URL.Path, nil
+}
+
+// ServeUserPrincipal replies to discovery requests. It returns false if the
+// request wasn't handled.
+func ServeUserPrincipal(w http.ResponseWriter, r *http.Request, backend UserPrincipalBackend) bool {
```
emersion, line 283: I wonder if we should just take a struct instead of an interface here. For instance:

```go
func ServeUserPrincipal(w http.ResponseWriter, r *http.Request, options *ServeUserPrincipalOptions) bool

type ServeUserPrincipalOptions struct {
    Path string
    HomeSets map[string]HomeSet
}
```

Should be very cheap to compute for the backend I think?

server.go
```diff
@@ -234,3 +237,131 @@ func (b *backend) Move(r *http.Request, dest *internal.Href, overwrite bool) (cr
...
+
+// RelativeResourcePath returns the path of a resource without the prefix of
+// the home set that the resource is in. If the resource is not in any known
+// home set, the original request path is returned.
+func RelativeResourcePath(r *http.Request, backend UserPrincipalBackend) (string, error) {
+	homeSets, err := backend.ListUserHomeSets(r.Context())
+	if err != nil {
+		return "", err
+	}
+	paths := make([]string, 0, len(homeSets))
```
emersion, line 258: No need for an intermediate slice here -- can directly loop over the map

server.go
bitfehler: I think it _could_ be used elsewhere, though it wouldn't have to if it's inconvenient. I was actually at some point considering attaching such a struct to the context. Maybe even use a middleware to do this?

bitfehler: Actually, we also need the interface to pass on to the `internal.Handler`, which needs it (in general, not so much for this function), so I guess we might as well stick to it? I might also be misunderstanding what you are aiming for.

server.go
bitfehler: Derp. Fixed.

server.go
emersion: I guess what I'm trying to get to is that `UserPrincipalBackend` is a bit overkill for the purposes of `ServeUserPrincipal`. We don't actually need the complexity of an interface with multiple functions here, a simple `ServeUserPrincipalOptions` struct with the info in it would be enough.

That said, it's true that CardDAV and CalDAV backends also need to access this info, and embedding `UserPrincipalBackend` in their backends is a simple way to require that. If we use a struct instead of an interface, we'd have to add

```go
UserPrincipal(context.Context) *ServeUserPrincipalOptions
```

to the CardDAV and CalDAV backend interfaces. At which point the name `ServeUserPrincipalOptions` is maybe not a good name, maybe something else like `UserPrincipalOptions` or `UserPrincipalInfo` would be more fitting.

Just thinking out loud here, not sure what the best solution is, and I'm probably thinking too much as I often do. :P

server.go
```diff
@@ -234,3 +237,128 @@ func (b *backend) Move(r *http.Request, dest *internal.Href, overwrite bool) (cr
 	}
 	return created, err
 }
+
+type HomeSet interface {
```
emersion, line 241: I'm not super happy about this interface, because it's pretty easy to implement it externally outside of go-webdav, and implement it wrong by missing the fact that it needs to be XML-marshalable. To alleviate this, I've been thinking about making it impossible for external users to implement it. Usually this is done by adding an unexported func in the interface, e.g.

```go
type HomeSet interface {
    homeSet()
}
```

that way only the current package can implement it. But this wouldn't work here because the type is shared between our own packages, and the `carddav` package won't be able to implement the type.

Then I thought about a bunch of different solutions, like moving it to `internal/` and doing more tricks to ensure only us can implement it. But it's pretty ugly and not worth the trouble.

tl;dr I think it's fine and we just need to document the type here. Explain what it represents, explain that it needs to marshal to XML to describe the home set element, etc.

emersion, line 241: This interface looks like it won't be easily re-usable in the DAV client. It would be nicer to find a good name to reflect this.

caldav/server.go
```diff
@@ -24,7 +25,8 @@ type Backend interface {
 // Handler handles CalDAV HTTP requests. It can be used to create a CalDAV
 // server.
 type Handler struct {
-	Backend Backend
+	Backend              Backend
+	UserPrincipalBackend webdav.UserPrincipalBackend
```
emersion, line 29: We could just embed `webdav.UserPrincipalBackend` into `Backend` to avoid dealing with multiple backends.

server.go
```diff
@@ -234,3 +237,128 @@ func (b *backend) Move(r *http.Request, dest *internal.Href, overwrite bool) (cr
...
+
+type UserPrincipalBackend interface {
+	CurrentUserPrincipal(ctx context.Context) (string, error)
+	ListUserHomeSets(ctx context.Context) (map[string]HomeSet, error)
+}
+
+// RelativeResourcePath returns the path of a resource without the prefix of
+// the home set that the resource is in. If the resource is not in any known
+// home set, the original request path is returned.
+func RelativeResourcePath(r *http.Request, backend UserPrincipalBackend) (string, error) {
```
emersion, line 253: Can we move `RelativeResourcePath` and `IsResourceRequest` to `internal/`? Or do we have good reasons to expose them?

bitfehler:
Ok, I think I addressed all of your comments. I also shuffled the interfaces a bit (user principal backend is only responsible for user principal URL, carddav/caldav backend are responsible for home set path). I think the implementation is reasonably clean, given the complexity of the matter. The application implementation would also be pretty straightforward.
I am not sure about the home set interface naming, I came up with `BackendSuppliedHomeSet`, but I'll use whatever else if you have a suggestion.

If you give a thumbs up I am also happy to add some more documentation.

server.go
```diff
@@ -234,3 +235,96 @@ func (b *backend) Move(r *http.Request, dest *internal.Href, overwrite bool) (cr
...
+
+type ServeUserPrincipalOptions struct {
+	UserPrincipalPath string
+	HomeSets          []BackendSuppliedHomeSet
+}
+
+// ServeUserPrincipal replies to discovery requests. It returns false if the
+// request wasn't handled.
+func ServeUserPrincipal(w http.ResponseWriter, r *http.Request, options ServeUserPrincipalOptions) bool {
+	if r.URL.Path == "/.well-known/carddav" || r.URL.Path == "/.well-known/caldav" {
```
emersion, line 262: Hm. I'm not a fan of hardcoding carddav and caldav here. Can we leave that to the appropriate packages?

server.go
```diff
@@ -234,3 +235,96 @@ func (b *backend) Move(r *http.Request, dest *internal.Href, overwrite bool) (cr
...
+		for _, homeSet := range b.options.HomeSets {
+			hs := homeSet // capture variable for closure
+			props[homeSet.GetXMLName()] = func(*internal.RawXMLValue) (interface{}, error) {
+				return hs, nil
+			}
+		}
+	}
+
+	resp, err := internal.NewPropfindResponse(r.URL.Path, propfind, props)
+	return internal.NewMultistatus(*resp), err
```
emersion, line 297: `NewPropfindResponse` can return a nil `resp` on error, in which case this panics.

carddav/server.go
```diff
@@ -406,6 +454,14 @@ func (b *backend) Proppatch(r *http.Request, update *internal.Propertyupdate) (*
 }
 
 func (b *backend) Put(r *http.Request) (*internal.Href, error) {
+	homeSetPath, err := b.Backend.AddressbookHomeSetPath(r.Context())
+	if err != nil {
+		return nil, err
+	}
+	if !strings.HasPrefix(r.URL.Path, homeSetPath) {
+		return nil, &internal.HTTPError{Code: http.StatusMethodNotAllowed}
+	}
```
emersion, line 463: I'm not sure it's a good idea to add this logic. Servers may want to allow users to fetch/edit calendars which aren't under the home set. Additionally, `strings.HasPrefix` is not appropriate here -- the check passes with the path `/abcdef` and the home set `/abc`.

carddav/server.go
```diff
@@ -330,6 +374,18 @@ func (b *backend) Propfind(r *http.Request, propfind *internal.Propfind, depth i
 	return internal.NewMultistatus(resps...), nil
 }
 
+func (b *backend) propfindCommon(propfind *internal.Propfind, path, principalPath, homeSetPath string) (*internal.Response, error) {
+	props := map[xml.Name]internal.PropfindFunc{
+		addressBookHomeSetName: func(*internal.RawXMLValue) (interface{}, error) {
```
emersion, line 379: This only needs to be defined on the user principal.

carddav/server.go
bitfehler: Ok, so a few things :slightly_smiling_face: 

> Servers may want to allow users to fetch/edit calendars which aren't under the home set

I agree that this won't work like this, but as things are right now, this is outright impossible anyways (due to the home set being `/`). I think part of why this PR turned a bit tedious might be that I cannot quite seem to gauge your expectations. I initially had the impression you'd actually prefer to only do the bare minimum required to get the two servers working side-by-side, which would have just been allowing a home set prefix. I guess it is on me that I brought the user principal URL into this. 

However, with this comment, I get the impression that now we're talking about laying the foundations for all kinds of things that have been previously impossible, such as accessing other people's calendars. This is fine, I just need to understand where we want to go.

Through #70 , I also realized that the backends should behave more like a HTTP backend (as opposed to e.g. map lookup or such). As such, my suggestion would be:

* we remove that check entirely
* as discussed in #70, we expose some appropriate internal errors (not found, access denied, etc)
* we document that the backend can have absolutely zero expectations w/ regards to the path it is asked about and has to check it itself and return the appropriate errors

WDYT?

PS: on a side note, I am quite disappointed that Go doesn't have a proper `path.HasPrefix()`, which is why I initially wrote those custom functions. But the check really would have never been good enough anyways, and the backend will have to perform them. So just removing it from here really makes sense...

emersion: > However, with this comment, I get the impression that now we're talking about laying the foundations for all kinds of things that have been previously impossible, such as accessing other people's calendars. This is fine, I just need to understand where we want to go.

From my PoV, I'm just trying not to add any more assumptions which we will regret later on. The `HasPrefix` checks are new here, and I don't think it's something we want to have long-term.

> Through https://github.com/emersion/go-webdav/pull/70 , I also realized that the backends should behave more like a HTTP backend (as opposed to e.g. map lookup or such).

Right. It's mostly about being able to return more detailed errors than "something failed". Just like `os.Open` errors can be introspected for instance.

> WDYT?

LGTM!

> I am quite disappointed that Go doesn't have a proper path.HasPrefix(), which is why I initially wrote those custom functions.

Yeah, I agree this is disappointing, as this is a common thing to do and it can be error-prone. Something like this can be used in backends:

```go
func pathHasPrefix(p, prefix string) bool {
    p = path.Clean(p)
    prefix = path.Clean(prefix)
    return p == prefix || strings.HasPrefix(p, prefix + "/")
}
```

bitfehler: Removed the checks as discussed and also addressed your other comments.

server.go
```diff
@@ -234,3 +235,94 @@ func (b *backend) Move(r *http.Request, dest *internal.Href, overwrite bool) (cr
...
+
+type ServeUserPrincipalOptions struct {
+	UserPrincipalPath string
+	HomeSets          []BackendSuppliedHomeSet
+}
+
+// ServeUserPrincipal replies to discovery requests. It returns false if the
+// request wasn't handled.
+func ServeUserPrincipal(w http.ResponseWriter, r *http.Request, options ServeUserPrincipalOptions) bool {
+	if r.Method != "PROPFIND" {
```
emersion, line 262: Hmm. If `r.URL.Path != b.options.UserPrincipalPath` we shouldn't handle the request, right? Should we return false here in that case?

emersion: (In which case, maybe we can just let the caller perform the path check and drop `UserPrincipalPath`?)

caldav/server.go
```diff
@@ -209,8 +227,30 @@ func (b *backend) HeadGet(w http.ResponseWriter, r *http.Request) error {
...
+	}
+	principalPath, err := b.Backend.CurrentUserPrincipal(r.Context())
+	if err != nil {
+		return nil, err
+	}
+
 	var resps []internal.Response
-	if r.URL.Path == "/" {
+
+	resp, err := b.propfindCommon(propfind, r.URL.Path, principalPath)
```
emersion, line 241: Hmm. This is a bit weird. It produces multiple `<response>` elements for the same href potentially.

server.go
bitfehler: I thought the idea was that this handles everything up to, including the user principal URL. I've observed at least Evolution sending PROPSTAT to `/` during discovery, which then returns the user principal URL.

bitfehler: That's also why the .well-know handling was in there.

caldav/server.go
bitfehler: Yeah, sorry, I think I refactored that once too often. The composability is a bit weird there. Are you okay with how it is now?

I think in the long run, if we add more basic WebDAV properties, it might make sense to pass around pre-filled maps. But I tried that, and it would require a map copy in some places, so it's a bit ugly also. For now, it seemed best to just follow the current pattern but add the additional property everywhere...

emersion: > Are you okay with how it is now?

Yeah sounds fine, let's keep it simple. Though, CardDAV seems fixed but not CalDAV.

server.go
emersion: Hm I'd rather just handle the principal stuff in here (as the function name suggests).

caldav/server.go
bitfehler: Sorry, that got lost in rebase-land :confused: fixed now

server.go
bitfehler: Ok, I rewrote that function based on #74 (feel free to add yourself as co-author :slightly_smiling_face: ). I didn't take everything from there to not blow this up any more, I hope that's fine. Happy to add more bits and pieces afterwards.

bitfehler:
For what it's worth, you can see how the usage of all this could look like in its current state here: https://git.sr.ht/~sircmpwn/tokidoki/tree/master/item/cmd/tokidoki/main.go

Note that I am using more stuff that I wrote that I have not yet created PRs for, as it would conflict with this one, you can find that here: https://github.com/bitfehler/go-webdav/commits/bitfehler/more-caldav I am constantly rebasing that branch on the latest developments here, so should be easy to merge the rest once we arrive at a solution.

server.go
```diff
@@ -234,3 +236,73 @@ func (b *backend) Move(r *http.Request, dest *internal.Href, overwrite bool) (cr
...
+		http.Error(w, "unsupported method", http.StatusMethodNotAllowed)
+	}
+}
+
+func serveUserPrincipalPropfind(w http.ResponseWriter, r *http.Request, options *ServeUserPrincipalOptions) error {
+	var propfind internal.Propfind
+	if err := internal.DecodeXMLRequest(r, &propfind); err != nil {
+		return err
+	}
+	props := map[xml.Name]internal.PropfindFunc{
```
emersion, line 283: We're missing a few principal props marked in the RFC as "MUST support", but we were missing them prior to this PR too. Let's keep that in mind for the future.

emersion:
Thanks!

server.go
badereddinkhmila: hey guys, i'm using this lib to serve carddav and i wonder why the restrictions on the urls !! is it defined in the rfc or what ?
Cause it's making it a bit harder to integrate this to an already developped application.
Thanks both of you.

emersion: Please open an issue with more details. I don't really understand what you're talking about here.

badereddinkhmila: okey i'll delete this comment and open an issue

PR #64, opened 2022-03-31
caldav: extend query filter types

bitfehler:
The basic types related to queries and filtering are missing some
features specified in the RFC (as also noted in the TODO comments). This
adds several of the missing elements, working towards being able to
handle all RFC-compliant queries.

The work is not fully done, e.g. the collation for text-match is still
not handled, but it's getting pretty close.


emersion (approved):
LGTM, thanks!

PR #66, opened 2022-04-11
Support custom user principal and home set paths (v2)

bitfehler:
Currently, the user principal path and the home set path are both
hardcoded to "/", for both CalDAV and CardDAV. This poses a challenge if
one wishes to run a CardDAV and CalDAV server in the same server.

This commit adds a bit plumbing to offer a convenience function,
`ServeListHomeSets`, that can serve PROPFIND responses listing both
CalDAV and CardDAV home-set properties (calendar-home-set and
addressbook-home-set respectively).

To support this, the server must be aware of the current user's
principal URL. This is not a concern of either CalDAV or CardDAV, but
WebDAV in general. This commit allows an implementation to supply this
URL through the request context. If not set, the previous convention of
using "/" remains in place.

The individual servers will continue to work as before (including the
option of keeping everything at "/"). If one wishes to run CardDAV and
CalDAV in parallel, the new `webdav.ServeListHomeSets()` can be used to
serve requests for the current user's principal URL. The request routing
is the implementation's responsibility.

For added convenience, the `ServeListHomeSets()` function includes
support for discovering the home sets via their `/.well-known` URLs.

The home set path has to be provided by the storage backend. It is
customary, though not required, that the home set path includes the
user's pricipal path. The storage backends already have access to the
request context, which allows computing the home set path based on the
pricipal path.

### DIFF
--- a/caldav/caldav.go
+++ b/caldav/caldav.go
@@ -7,7 +7,13 @@ import (
 	"time"
 
 	"github.com/emersion/go-ical"
+	"github.com/emersion/go-webdav"
+	"github.com/emersion/go-webdav/internal"
 )
+
+func NewCalendarHomeSet(path string) webdav.ListableHomeSet {
+	return &calendarHomeSet{Href: internal.Href{Path: path}}
+}
 
 type Calendar struct {
 	Path            string
--- a/caldav/elements.go
+++ b/caldav/elements.go
@@ -30,6 +30,10 @@ type calendarHomeSet struct {
 	Href    internal.Href `xml:"DAV: href"`
 }
 
+func (a *calendarHomeSet) GetXMLName() xml.Name {
+	return calendarHomeSetName
+}
+
 // https://tools.ietf.org/html/rfc4791#section-5.2.1
 type calendarDescription struct {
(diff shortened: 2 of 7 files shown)


bitfehler:
I'd like to offer this as an alternative to #62. It incorporates most of your comments, plus one other thing: I think the user principal URL should be decoupled from the rest (which is done here by using the context to communicate it). The CalDAV/CardDAV backends have to be able to relate a request path to one specific path: the home-set. While usually the case, there is no rule that a home-set has to be a sub-path of the user's principal URL. A backend can decide to do so or not.

As such, I think using the context for this makes the most sense. Everything else I tried to simplify as much as possible, mostly per your comments on #62. Note that 

> This interface looks like it won't be easily re-usable in the DAV client. It would be nicer to find a good name to reflect this.

is a pretty tough challenge. Let me know what you think, but I am also open for suggestions.

If you don't like this, that's cool, I can go ahead and simplify #62 accordingly, but I thought it's worth discussing.

bitfehler:
Forgot to mention: I think the user-principal-url is more of a concern of "user management", possibly authentication or such.

bitfehler:
As discussed on IRC, the context may not have been the best approach and it's better to keep things explicit. Closing this in favor of refining #62

PR #67, opened 2022-04-12
caldav: add supported-calendar-component-set field

koukas:
Applications can detect the type of a calendar collection based on the `supported-calendar-component-set` property. This adds a respective field to the `Calendar` struct.


emersion (approved):
LGTM, thanks!

Issue #50, opened 2021-07-17
CalDav ETag UnmarshalText issue


skroczek:
I experimented a bit with this library and found an issue with caldavClient.QueryCalendar.
I am trying to query a calendar from OpenXchange but I get the error "webdav: failed to unquote ETag: invalid syntax". This error seems to come from the file ["internal/elements.go:345"](https://github.com/emersion/go-webdav/blob/8efde26ef9fd73acccd60b443e4bb7c02da5e6f0/internal/elements.go#L345) and occurs because in the returned event the ETag looks like this: "<D:getetag>http://www.open-xchange.com/etags/6063-1626334223127</D:getetag>".
Unfortunately, I don't know how to fix this. Should you need more information, I will of course make it available.

Program source
```go
package main

import (
	"fmt"
	"github.com/emersion/go-webdav"
	"github.com/emersion/go-webdav/caldav"
	"log"
	"net/http"
	"net/http/httptrace"
	"net/http/httputil"
	"os"
	"time"
)

// transport is an http.RoundTripper that keeps track of the in-flight
// request and implements hooks to report HTTP tracing events.
type transport struct {
	current *http.Request
}

// RoundTrip wraps http.DefaultTransport.RoundTrip to keep track
// of the current request.
func (t *transport) RoundTrip(req *http.Request) (*http.Response, error) {
	t.current = req
	fmt.Println(req)
	resp, err := http.DefaultTransport.RoundTrip(req)
	dump, _ := httputil.DumpResponse(resp, true)

	fmt.Printf("\n\nResponse:\n%q\n", dump)
	return resp, err
}

func main() {
	t := &transport{}

	client := &http.Client{Transport: t}

	baHttpClient := webdav.HTTPClientWithBasicAuth(client, os.Getenv("USERNAME1"), os.Getenv("PASSWORD"))

	caldavClient, err := caldav.NewClient(baHttpClient, os.Getenv("ENDPOINT"))
	if err != nil {
		log.Fatal(err)
	}

	principal, err := caldavClient.FindCurrentUserPrincipal()
	if err != nil {
		log.Fatal(err)
	}

	fmt.Println(principal)
	homeSet, err := caldavClient.FindCalendarHomeSet(principal)
	if err != nil {
		log.Fatal(err)
	}
	fmt.Println(homeSet)
	calendars, err := caldavClient.FindCalendars(homeSet)
	if err != nil {
		log.Fatal(err)
	}
	fmt.Println(calendars)

[...]

PR #69, opened 2022-04-18
Allow parse etag without quotes

LugaMuga:
Catched issue same like https://github.com/emersion/go-webdav/issues/50.
Understand that it's needs when server not working strictly by RFC. But some developers will try to use library for "wrong implemented" servers and get the same error.
Think it's possible to make requirement not so strictly in that case.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,8 @@
 
 # Project-local glide cache, RE: https://github.com/Masterminds/glide/issues/736
 .glide/
+
+# JetBrains GoLand
+.idea
+*.iml
+*.ipr
--- a/internal/elements.go
+++ b/internal/elements.go
@@ -342,7 +342,8 @@ type ETag string
 func (etag *ETag) UnmarshalText(b []byte) error {
 	s, err := strconv.Unquote(string(b))
 	if err != nil {
-		return fmt.Errorf("webdav: failed to unquote ETag: %v", err)
+		*etag = ETag(b)
+		return nil
 	}
 	*etag = ETag(s)
 	return nil


.gitignore
```diff
@@ -12,3 +12,8 @@
 
 # Project-local glide cache, RE: https://github.com/Masterminds/glide/issues/736
 .glide/
+
+# JetBrains GoLand
+.idea
+*.iml
+*.ipr
```
emersion, line 19: Please don't commit editor-specific ignore rules.

internal/elements.go
```diff
@@ -342,7 +342,8 @@ type ETag string
 func (etag *ETag) UnmarshalText(b []byte) error {
 	s, err := strconv.Unquote(string(b))
 	if err != nil {
-		return fmt.Errorf("webdav: failed to unquote ETag: %v", err)
+		*etag = ETag(b)
```
emersion, line 345: Hm, this completely ignores all errors. Not sure this is a great idea.

emersion:
TBH, maybe we shouldn't do any of the quoting/unquoting stuff. It would be a bit awkward to deal with quoted strings, but at least we wouldn't have this kind of issues and we would handle weak entity tags properly.

internal/elements.go
LugaMuga: Agreed. But how we can logged error with warn status?

LugaMuga:
Agree. But what will happen with projects, that already use this library? 
I don't think that they have tests for quotes, and this change can broke them.

emersion:
We've already made plenty of breaking changes since the last release.

LugaMuga:
ok, changed patch as you said

internal/elements.go
```diff
@@ -340,11 +340,7 @@ type GetETag struct {
 type ETag string
 
 func (etag *ETag) UnmarshalText(b []byte) error {
```
emersion, line 342: This method is now a no-op and can be dropped.

emersion (changes requested):
`String` and `MarshalText` should be dropped as well, otherwise ETags will get double-quoted.

LugaMuga:
> `String` and `MarshalText` should be dropped as well, otherwise ETags will get double-quoted.

Needs String method when you write header with eTag

internal/elements.go
LugaMuga: ok

.gitignore
```diff
@@ -11,4 +11,4 @@
 *.out
 
 # Project-local glide cache, RE: https://github.com/Masterminds/glide/issues/736
-.glide/
```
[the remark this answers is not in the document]
LugaMuga: reverted all changes

internal/elements.go
LugaMuga: done

emersion:
Hm, this new proposal is error-prone. Library users need to "guess" that they need to double-quote the ETag values they're using. It's very easy to break the RFC with this API.

LugaMuga:
That's why I worried about remove String method. 
Let's I return it and in server call internal.ETag(ao.ETag).String() ?

emersion:
`string(ao.ETag)` should work as well.

LugaMuga:
Nope. String() method, that I was removed, is adding quotes to ETag value. 
This is not the same string(ao.ETag)

emersion:
With your changes, the ETag contains quotes already, no need to add them again.

LugaMuga:
Where my changes add quotes in code ?
`w.Header().Set("ETag", ao.ETag)`

LugaMuga:
Removed wrapper, with call String() method on it. I think it will broke server in setting ETag to headers

LugaMuga:
Could you please explain, why need write type conversion here? 

Screenshot 2022-06-14 at 20 57 13

PR #71, opened 2022-04-29
Small error improvements

emersion:
cc @bitfehler

### DIFF
--- a/caldav/server.go
+++ b/caldav/server.go
@@ -327,14 +327,10 @@ const (
 func NewPreconditionError(err PreconditionType) error {
 	name := xml.Name{"urn:ietf:params:xml:ns:caldav", string(err)}
 	elem := internal.NewRawXMLElement(name, nil, nil)
-	e := internal.Error{
-		Raw: []internal.RawXMLValue{
-			*elem,
-		},
-	}
-	return &internal.DAVError{
+	return &internal.HTTPError{
 		Code: 409,
-		Msg:  fmt.Sprintf("precondition not met: %s", string(err)),
-		Err:  e,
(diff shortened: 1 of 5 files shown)


emersion:
Hm, "internal: return a DAVError instead of an Error" is a bit strange because it wraps `DAVError` into `HTTPError`, duplicating the HTTP code. We might want to pick between two strategies:

- Use `HTTPError` with an `Error` wrapped in it to describe an XML DAV error. Remove `DAVError`.
- Use `DAVError` to describe an XML DAV error, and use `HTTPError` for HTTP errors without an XML description. Never use `Error` as a Go error value.

bitfehler:
Sorry, I never really looked much at the client code before, and it took me a moment to understand the error handling you added there :slightly_smiling_face: 

I am not entirely sure which approach sounds better. On the one hand, since both are essentially different protocols, separating them seems reasonable. Yet, since they map so closely to each other, maybe not. I think in the client library, both would be easy to implement, but I have no idea how tedious proper handling would be in a client application.

On the server side, however, I'd be worried that separating them would result in a lot of noisy code, like "if it's a http error and code X or a dav error and code X, do this, same for code Y, etc." without adding much value. If the only difference is a human-readable error message, getting that from an optional wrapped error seems much easier. Hence I'd slightly lean toward your first suggestion. But it's not a strong opinion and I might very well be missing something.

emersion:
Here's how the first approach would look like.

bitfehler:
Looks good to me... :slightly_smiling_face:

emersion:
Thanks for the feedback!

PR #73, opened 2022-05-03
Add exported function to create `HTTPError`

bitfehler:
This can be used by backends to influence the status code returned to
clients for errors that occurred in the backend.


bitfehler:
As discussed on IRC. For now, I don't see the need to expose this type at all. Maybe the client code could wrap (or convert) this (and potentially other errors it might encounter) into its own type and export that instead?

emersion:
Thanks!

PR #76, opened 2022-05-12
Implement GET/HEAD/PUT for calendar objects

bitfehler:
Still missing a few features, but works to a certain extend. Requires an update to the backend interface to support the operations.


emersion (approved):
LGTM, thanks!

PR #77, opened 2022-05-12
webdav: rename ServeUserPrincipal to ServePrincipal

emersion:
A principal may represent something else than a user, for instance
it may represent a group.

Also rename UserPrincipalPath to CurrentUserPrincipalPath, because
the principal being served may not represent the current user.

cc @bitfehler


bitfehler (approved):
Good point, I totally agree!

PR #80, opened 2022-05-12
caldav: implement filter function for queries

bitfehler:
This is not yet complete (see TODOs in code), but basic filtering of a
list of CaledarObjects works.

Includes test data from the RFC, which allows to use the RFCs examples
as test cases.


bitfehler:
Sorry for bombarding you with PRs, this is all I have for now :slightly_smiling_face:

bitfehler:
Noticed I hadn't run `go fmt` on this, sorry about that.

caldav/match.go
```diff
@@ -0,0 +1,224 @@
...
+
+// Filter returns the filtered list of calendar objects matching the provided query.
+// A nil query will return the full list of calendar objects.
+func Filter(query *CalendarQuery, cos []CalendarObject) ([]CalendarObject, error) {
+	if query == nil {
+		// FIXME: should we always return a copy of the provided slice?
+		return cos, nil
+	}
+
+	out := make([]CalendarObject, 0)
```
emersion, line 17: Style nit: either specify the capacity (e.g. `len(cos)`), either switch to `var out []CalendarObject`.

caldav/match.go
```diff
@@ -0,0 +1,224 @@
...
+		// TODO properties are not currently filtered even if requested
+		out = append(out, co)
+	}
+	return out, nil
+}
+
+// Match reports whether the provided CalendarObject matches the query.
+func Match(query CompFilter, co *CalendarObject) (matched bool, err error) {
+	if co.Data == nil || co.Data.Component == nil {
+		// TODO is this a thing? Should this match if comp.IsNotDefined?
```
emersion, line 36: It's fine to crash/panic in this case, since that would be a backend bug.

caldav/match.go
```diff
@@ -0,0 +1,224 @@
...
+		return false, nil
+	}
+	comp := co.Data.Component
+
+	if query.Name != comp.Name {
```
emersion, line 41: According to https://datatracker.ietf.org/doc/html/rfc4791#section-9.7.1, `query.Name` may match a child of the calendar object. `VCALENDAR` is the root component, but the request might match a child VEVENT component.

emersion: Hm, in that VEVENT case it seems like the child comp-filter must be enclosed in a parent VCALENDAR comp-filter, so should be fine.

caldav/match.go
```diff
@@ -0,0 +1,224 @@
...
+			matches = append(matches, child)
+		}
+	}
+	if len(matches) == 0 {
+		return filter.IsNotDefined, nil
+	}
+	return true, nil
+}
+
+func matchCompFilterChild(filter CompFilter, comp *ical.Component) (bool, error) {
```
emersion, line 83: Hm. It seems like this duplicates code with `Match`. Can't we remove the root component special case?

caldav/match.go
```diff
@@ -0,0 +1,224 @@
...
+		}
+		if !match {
+			return false, nil
+		}
+	}
+	return true, nil
+}
+
+func matchPropFilter(filter PropFilter, comp *ical.Component) (bool, error) {
+	// TODO: this only matches first field, can there be multiple like for CardDAV?
```
emersion, line 120: Yeah, there could be multiple

caldav/match.go
```diff
@@ -0,0 +1,224 @@
...
+	}
+	if filter.TextMatch != nil {
+		return matchTextMatch(*filter.TextMatch, value)
+	}
+	return true
+}
+
+func matchTextMatch(txt TextMatch, value string) bool {
+	// TODO: handle text-match collation attribute
+	match := value == txt.Text
```
emersion, line 219: According to https://datatracker.ietf.org/doc/html/rfc4791#section-9.7.5, text-match specifies as substring match.

caldav/match.go
```diff
@@ -0,0 +1,224 @@
...
+	var zeroDate time.Time
+	if filter.Start != zeroDate {
+		match, err := matchPropTimeRange(filter.Start, filter.End, field)
+		if err != nil {
+			return false, err
+		}
+		if !match {
+			return false, nil
+		}
+		for _, paramFilter := range filter.ParamFilter {
```
emersion, line 135: Since this loop is common to both `if` branches, maybe move it one scope level up?

caldav/match.go
bitfehler: Thanks for nudging me on this, the recursion is much cleaner now.

caldav/match.go
bitfehler: Oops, thanks, changed and added a test case for that.

caldav/match.go
bitfehler: I moved it up. But: according to [RFC 4791, section 9.7.2](https://datatracker.ietf.org/doc/html/rfc4791#section-9.7.2), the param filters are only present/evaluated if one of the if-branches is taken. I suppose it doesn't matter, because for any valid message, there is no semantic difference. I just think the code reads a little odd when compared to the RFC.

caldav/match.go
```diff
@@ -0,0 +1,224 @@
...
+			}
+		}
+		return true, nil
+	}
+	// empty prop-filter, property exists
+	return true, nil
+}
+
+func matchCompTimeRange(start, end time.Time, comp *ical.Component) (bool, error) {
+	// TODO what about other types of components?
```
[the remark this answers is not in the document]
bitfehler: I did. I think the time range matching will need some more lovin', but I still think it's a start...

emersion (approved):
This version LGTM, thanks!

PR #81, opened 2022-05-12
Only call CurrentUserPrincipal when necessary

emersion:
Very minor cleanup to avoid unnecessarily calling `CurrentUserPrincipal`.


bitfehler (approved):
:shipit:

PR #83, opened 2022-05-17
caldav: add support for getcontentlength property

bitfehler:
Allow the backend to provide a value for the `getcontentlength` property
as described in [RFC 2518 section 13.4][1].

The implementation treats is as optional, allthough it is a required
property per RFC. Most clients do perfectly fine without it, though.

Properly setting this in the backend makes the CalDAV collection
listable with clients that do require it, e.g. cadaver.

[1]: https://datatracker.ietf.org/doc/html/rfc2518#section-13.4

### DIFF
--- a/caldav/caldav.go
+++ b/caldav/caldav.go
@@ -117,8 +117,9 @@ type CalendarMultiGet struct {
 }
 
 type CalendarObject struct {
-	Path    string
-	ModTime time.Time
-	ETag    string
-	Data    *ical.Calendar
+	Path          string
+	ModTime       time.Time
+	ContentLength int64
+	ETag          string
+	Data          *ical.Calendar
 }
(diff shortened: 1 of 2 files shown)


emersion:
Can we also set the `Content-Length` header in `HeadGet`?

bitfehler:
Hm, interesting. I didn't do that because in my tests it was always already there. But I just read the docs for [ResponseWriter](https://pkg.go.dev/net/http#ResponseWriter), and they say

> Additionally, if the total size of all written data is under a few KB and there are no Flush calls, the Content-Length header is added automatically.

So I assume that is not to be relied upon and we should do it explicitly?

emersion:
Hm, that's interesting indeed. Yeah, the body could contain a lot of data with embedded resources (contact pictures, etc), so it's better to set it manually if the backend supplies it.

emersion (approved):
Thanks!

PR #84, opened 2022-05-17
caldav: add support for getcontentlength property

bitfehler:
Allow the backend to provide a value for the `getcontentlength` property
as described in [RFC 2518 section 13.4][1].

The implementation treats is as optional, allthough it is a required
property per RFC. Most clients do perfectly fine without it, though.

Properly setting this in the backend makes the CalDAV collection
listable with clients that do require it, e.g. cadaver.

[1]: https://datatracker.ietf.org/doc/html/rfc2518#section-13.4


emersion:
If you get the chance, can you backport the changes from the other MR here as well? Otherwise I can do it as well.

bitfehler:
Done, including the client part

PR #86, opened 2022-05-19
carddav: do property filtering in match.Filter() 

bitfehler:
I am a bit unsure whether or not this makes sense and would like to solicit your input :slightly_smiling_face: 

I would say that this implementation makes the behavior of the `Filter()` function more correct. However, a case could be made that this performs work that e.g. backend may have already performed. Example: a hypothetical database backend might construct a database query that already returns the objects reduced to the requested properties (and/or only those matching the query). In this case, the current implementation would simply create a copy of each object with no further benefit.

However, I would argue that that is a very advanced use case, and such a backend would probably not be using the `Filter()` function in the first place. Also, while I could imagine such a hypothetical backend for carddav, the situation for caldav is much more complex (recursive data structures), and I think having a correct reference implementation at all would very much outweigh the performance considerations here.

Let me know how you feel about this. Also, regardless of outcome, the first commit should be merged, but I am happy to create another PR for that if we decide that this is not fit.

### DIFF
--- a/carddav/match.go
+++ b/carddav/match.go
@@ -7,6 +7,34 @@ import (
 	"github.com/emersion/go-vcard"
 )
 
+func filterProperties(req AddressDataRequest, ao AddressObject) AddressObject {
+	if req.AllProp || len(req.Props) == 0 {
+		return ao
+	}
+
+	if len(ao.Card) == 0 {
+		panic("request to process empty vCard")
+	}
+
+	result := AddressObject{
+		Path:    ao.Path,
+		ModTime: ao.ModTime,
+		ETag:    ao.ETag,
+	}
+
+	result.Card = make(vcard.Card)
+	// result would be invalid w/o version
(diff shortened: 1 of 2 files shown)


emersion:
Yes, I think performing filtering by default makes sense. We can always expose more low-level functions later on when the need arises.

carddav/match.go
```diff
@@ -7,6 +7,35 @@ import (
 	"github.com/emersion/go-vcard"
 )
 
+func FilterProperties(req AddressDataRequest, ao AddressObject) AddressObject {
```
emersion, line 10: I'd prefer not to expose this function for now.

carddav/match.go
```diff
@@ -7,6 +7,35 @@ import (
...
+	if req.AllProp || len(req.Props) == 0 {
+		return ao
+	}
+
+	// TODO can this happen?
```
emersion, line 15: That would be a backend bug.

emersion (approved):
LGTM, thanks!

PR #88, opened 2022-05-20
client: support redirects in FindCurrentUserPrincipal

bitfehler:
One method to find the current user principal URL is to request the
`/.well-known/carddav` URL (see [RFC 6764, section 6][1]).

Add this as second attempt if the discovery via the configured endpoint
fails. This helps the client discover resources even if just given an
endpoint without path.

[1]: https://datatracker.ietf.org/doc/html/rfc6764#section-6

### DIFF
--- a/internal/client.go
+++ b/internal/client.go
@@ -149,7 +149,11 @@ func (c *Client) PropfindFlat(path string, propfind *Propfind) (*Response, error
 		return nil, err
 	}
 
-	return ms.Get(c.ResolveHref(path).Path)
+	// If the client followed a redirect, the Href might be different from the request path
+	if len(ms.Responses) != 1 {
+		return nil, fmt.Errorf("PROPFIND with Depth: 0 returned %d responses", len(ms.Responses))
+	}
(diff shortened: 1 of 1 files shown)


bitfehler:
Derp, I am sorry, you are absolutely right. I didn't even realize that this is in the `webdav` package. I'll provide a better solution later

bitfehler:
Sorry, I was actually conflating two things here. I updated this to only allow `FindCurrentUserPrincipal()` to follow redirects. Regardless of whether a `/.well-known` URL is the result of some discovery mechanism or provided explicitly by the user, the expectation is that a client can work based off of that URL. This would now work. We can tackle the actually discovery seperately.

emersion:
So this removes the path check `PropfindFlat` has.

I wonder, do you see the new path in `http.Client.CheckRedirect`? If so, we could implement a more robust `PropfindFlat` which takes HTTP redirections into account.

bitfehler:
As discussed on IRC, is this what you had in mind?

emersion (approved):
Thanks!

PR #94, opened 2022-05-31
internal: fix XML element struct naming

emersion:
This has been bugging me for a while. We were sometimes using TitleCase, sometimes Lowercase. Let's align on the idiomatic Go naming and pick TitleCase everywhere.

cc @bitfehler

### DIFF
--- a/caldav/client.go
+++ b/caldav/client.go
@@ -35,8 +35,8 @@ func NewClient(c webdav.HTTPClient, endpoint string) (*Client, error) {
 }
 
 func (c *Client) FindCalendarHomeSet(principal string) (string, error) {
-	propfind := internal.NewPropNamePropfind(calendarHomeSetName)
-	resp, err := c.ic.PropfindFlat(principal, propfind)
+	propfind := internal.NewPropNamePropFind(calendarHomeSetName)
+	resp, err := c.ic.PropFindFlat(principal, propfind)
 	if err != nil {
 		return "", err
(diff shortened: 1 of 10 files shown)


bitfehler:
Thanks, I was actually thinking about something like this myself the other day, but put it off for other things :slightly_smiling_face: 

Question, though: how do you determine what constitutes a "word" here? E.g. PropFind - I get that it's supposed to mean "Property Find", but it's sort of its own word, so don't have a clear intuition for this.

Let me give you an example of why I am asking: the example where I noticed this was in the carddav package. Sometimes it was `AddressBook` and sometimes `Addressbook`. I think the correct english is "address book", so it would be `AddressBook`. But then in many cases the name reflects XML elements, and they are for example `addressbook-description`, so I figured this would map more cleanly to `AddressbookDescription`. It seems, however, webdav itself has basically no dashes in elements at all ("responsedescription", "propertyupdate", ...) :confused: .

So, in summary: I am all for it, just want to make sure I understand the algorithm you are applying here. Is it based on "proper english" word splitting, regardless of XML names or such?

emersion:
Yeah, the WebDAV stuff usually doesn't make a lot of sense in general as you noticed and makes up for pretty weird-looking names. I'm just using the English words here, trying to make it as easy as possible for library users to parse the names.

emersion:
Thanks for the feedback!

PR #96, opened 2022-06-14
carddav: end-to-end test address book discovery

bitfehler:
As the implementation evolves, it will be necessary to have more tests
to assert we don't break anything when making changes. This commit
introduces a test setup to test that server and client can handle the
address book discovery with various parameters. The test setup should be
easily extendable to cover even more ground as needed.

### DIFF
--- a/carddav/carddav_test.go
+++ b/carddav/carddav_test.go
@@ -1,0 +1,174 @@
+package carddav
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	"github.com/emersion/go-vcard"
+	"github.com/emersion/go-webdav"
+)
+
+type testBackend struct{}
+
+type contextKey string
+
+var (
+	aliceData = `BEGIN:VCARD
+VERSION:4.0
+UID:urn:uuid:4fbe8971-0bc3-424c-9c26-36c3e1eff6b1
+FN;PID=1.1:Alice Gopher
+N:Gopher;Alice;;;
+EMAIL;PID=1.1:alice@example.com
+CLIENTPIDMAP:1;urn:uuid:53e374d9-337e-4727-8803-a1e9c14e0551
+END:VCARD`
+	alicePath = "urn:uuid:4fbe8971-0bc3-424c-9c26-36c3e1eff6b1.vcf"
+
+	currentUserPrincipalKey = "test:currentUserPrincipal"
+	homeSetPathKey          = "test:homeSetPath"
+	addressBookPathKey      = "test:addressBookPath"
+)
+
+func (*testBackend) CurrentUserPrincipal(ctx context.Context) (string, error) {
+	r := ctx.Value(currentUserPrincipalKey).(string)
+	return r, nil
+}
+
+func (*testBackend) AddressbookHomeSetPath(ctx context.Context) (string, error) {
+	r := ctx.Value(homeSetPathKey).(string)
+	return r, nil
+}
+
+func (*testBackend) AddressBook(ctx context.Context) (*AddressBook, error) {
+	p := ctx.Value(addressBookPathKey).(string)
+	return &AddressBook{
+		Path:                 p,
+		Name:                 "My contacts",
+		Description:          "Default address book",
+		MaxResourceSize:      1024,
+		SupportedAddressData: nil,
+	}, nil
+}
+
+func (*testBackend) GetAddressObject(ctx context.Context, path string, req *AddressDataRequest) (*AddressObject, error) {
+	if path == alicePath {
+		card, err := vcard.NewDecoder(strings.NewReader(aliceData)).Decode()
+		if err != nil {
+			return nil, err
+		}
+		return &AddressObject{
+			Path: path,
+			Card: card,
+		}, nil
+	} else {
+		return nil, webdav.NewHTTPError(404, fmt.Errorf("Not found"))
+	}
+}
+
+func (b *testBackend) ListAddressObjects(ctx context.Context, req *AddressDataRequest) ([]AddressObject, error) {
+	alice, err := b.GetAddressObject(ctx, alicePath, req)
+	if err != nil {
+		return nil, err
+	}
+
+	return []AddressObject{*alice}, nil
+}
+
+func (*testBackend) QueryAddressObjects(ctx context.Context, query *AddressBookQuery) ([]AddressObject, error) {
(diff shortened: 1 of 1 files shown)


bitfehler:
I created this mostly for #87 but I think it would be useful to have something like this for all sorts of reasons. I can of course do the same thing for carddav and even one for a combined server. But as usual, I wanted to get your feedback early on, as it usually proves to be very valuable :slightly_smiling_face: 

If you think it makes sense I would also extend the test function to at least list all contacts and explicitly retrieve one (well, the only) test contact.

carddav/carddav_test.go
```diff
@@ -0,0 +1,173 @@
...
+	alicePath = "urn:uuid:4fbe8971-0bc3-424c-9c26-36c3e1eff6b1.vcf"
+
+	currentUserPrincipalKey = "test:currentUserPrincipal"
+	homeSetPathKey          = "test:homeSetPath"
+	addressBookPathKey      = "test:addressBookPath"
+)
+
+func (*testBackend) CurrentUserPrincipal(ctx context.Context) (string, error) {
+	r, ok := ctx.Value(currentUserPrincipalKey).(string)
+	if !ok {
```
emersion, line 37: Nit: this is a test, we can just let it crash instead of having explicit error handling.

carddav/carddav_test.go
```diff
@@ -0,0 +1,173 @@
...
+func (*testBackend) PutAddressObject(ctx context.Context, path string, card vcard.Card, opts *PutAddressObjectOptions) (loc string, err error) {
+	panic("TODO: implement")
+}
+
+func (*testBackend) DeleteAddressObject(ctx context.Context, path string) error {
+	panic("TODO: implement")
+}
+
+func TestAddressBookDiscovery(t *testing.T) {
+
```
emersion, line 102: Style nit: stray newline

carddav/carddav_test.go
```diff
@@ -0,0 +1,173 @@
...
+		},
+		{
+			name:                 "explicit-well-known",
+			currentUserPrincipal: "/test/",
+			homeSetPath:          "/test/contacts/",
+			addressBookPath:      "/test/contacts/",
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+
```
emersion, line 130: Style nit: stray newline

carddav/carddav_test.go
```diff
@@ -0,0 +1,173 @@
...
+		//	addressBookPath: "/",
+		//},
+		{
+			name:                 "simple-home-set-path",
+			currentUserPrincipal: "/",
+			homeSetPath:          "/contacts/",
+			addressBookPath:      "/contacts/",
+		},
+		{
+			name:                 "explicit-well-known",
```
emersion, line 123: Hm, I'm not sure this is an accurate test case name, the CardDAV backend has no say in what the well-known URI is. Maybe "separate-principal" would be a better name?

carddav/carddav_test.go
```diff
@@ -0,0 +1,173 @@
...
+			currentUserPrincipal: "/test/",
+			homeSetPath:          "/test/contacts/",
+			addressBookPath:      "/test/contacts/",
+		},
```
emersion, line 127: Maybe we can have another test case with separate URIs for home-set and addressbook?

```go
		{
			name:                 "all-separate-paths",
			currentUserPrincipal: "/test",
			homeSetPath:          "/test/contacts",
			addressBookPath:      "/test/contacts/my-address-book",
		},
```

carddav/carddav_test.go
```diff
@@ -0,0 +1,173 @@
...
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+
+			h := Handler{&testBackend{}}
+			ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				r = r.WithContext(context.WithValue(r.Context(), currentUserPrincipalKey, tc.currentUserPrincipal))
+				r = r.WithContext(context.WithValue(r.Context(), homeSetPathKey, tc.homeSetPath))
+				r = r.WithContext(context.WithValue(r.Context(), addressBookPathKey, tc.addressBookPath))
```
emersion, line 135: We can omit repeated `r.WithContext` calls:

```go
ctx := r.Context()
ctx = context.WithValue(ctx, …)
ctx = context.WithValue(ctx, …)
ctx = context.WithValue(ctx, …)
r = r.WithContext(ctx)
```

carddav/carddav_test.go
```diff
@@ -0,0 +1,173 @@
...
+				(&h).ServeHTTP(w, r)
+			}))
+			defer ts.Close()
+
+			// client supports .well-known discovery if explicitly pointed to it
+			startURL := ts.URL
+			if tc.currentUserPrincipal != "/" {
+				startURL = ts.URL + "/.well-known/carddav"
```
emersion, line 143: Opened https://github.com/emersion/go-webdav/pull/98 to address this.

carddav/carddav_test.go
```diff
@@ -0,0 +1,173 @@
...
+			if hsp != tc.homeSetPath {
+				t.Fatalf("Found home set path '%s', expected '%s'", hsp, tc.homeSetPath)
+			}
+			abs, err := client.FindAddressBooks(hsp)
+			if err != nil {
+				t.Fatalf("error finding address books: %s", err)
+			}
+			if len(abs) != 1 {
+				t.Fatalf("Found %d address books, expected 1", len(abs))
+			}
```
emersion, line 170: Maybe check `abs[0].Path` here?

emersion (changes requested):

carddav/carddav_test.go
bitfehler: done

carddav/carddav_test.go
bitfehler: Agree, but "separate-principal" is also true for the previous case. I called it `nothing-at-root`, as this is the distinctive feature of this case. Does that work?

carddav/carddav_test.go
bitfehler: added

carddav/carddav_test.go
bitfehler: indeed, added. thanks :innocent:

carddav/carddav_test.go
emersion: Right!

emersion (approved):
Thanks!

PR #99, opened 2022-07-13
carddav: fix PROPFIND for all path layouts

bitfehler:
Refactor the PROPFIND handling to properly support all possible path
layouts (from "everything at /" to "everything at different path"). This
includes adding a new method the Backend interface, as it is the
authority that determines the path layout.

This fixes the recent regression of the broken "everything at /" layout.
It also already provides the needed bits and pieces to support multiple
address books.

### DIFF
--- a/carddav/server.go
+++ b/carddav/server.go
@@ -25,8 +25,18 @@ type PutAddressObjectOptions struct {
 	IfMatch string
 }
 
+type ResourceType int
+
+const (
+	ResourceTypeUserPrincipal      ResourceType = 1 << iota
+	ResourceTypeAddressBookHomeSet ResourceType = 1 << iota
+	ResourceTypeAddressBook        ResourceType = 1 << iota
+	ResourceTypeAddressObject      ResourceType = 1 << iota
+)
+
 // Backend is a CardDAV server backend.
 type Backend interface {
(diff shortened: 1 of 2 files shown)


carddav/server.go
```diff
@@ -25,8 +25,18 @@ type PutAddressObjectOptions struct {
...
+const (
+	ResourceTypeUserPrincipal      ResourceType = 1 << iota
+	ResourceTypeAddressBookHomeSet ResourceType = 1 << iota
+	ResourceTypeAddressBook        ResourceType = 1 << iota
+	ResourceTypeAddressObject      ResourceType = 1 << iota
+)
+
 // Backend is a CardDAV server backend.
 type Backend interface {
+	CardDAVResourceType(ctx context.Context, path string) (ResourceType, error)
```
bitfehler, line 39: One thing I was unsure about: I would have just named this `ResourceType()`, but we'd need the same thing for CalDAV eventually, and it's better to allow one type to implement both interfaces (i.e. keep the interfaces conflict-free).

One option might be to make this generic enough, e.g.: user principal, home set, collection, object - and then use that for both? Not sure, let me know if you have ideas...

bitfehler:
Made obsolete by #101

Issue #100, opened 2022-08-17
RFC: only support a single path layout?


bitfehler:
As came up several times, the code keeps getting more and more complicated due to the fact that we support different path layouts (as defined below). The latest example for this is #99. While that PR does a lot of the work already, the complexity of adding features (such as multiple address books might) will always be much higher when accounting for support for all use cases.

I figured it might be wise to take a step back and reiterate how we got here, what problems we are actually trying to solve, and if there is maybe different approach altogether that could ease the pain here.

Let's start with some definitions: WebDAV/CardDAV/CalDAV define a set of paths that serve certain special purposes. We need to consider the following:

* **User principal URL** - defined in [RFC 5397](https://datatracker.ietf.org/doc/html/rfc5397), identifies a user, actually an extension, but required by CardDAV/CalDAV. Also indispensable for discovery purposes, so we need this. Its relation to other paths is not specified, so it could be anything really. However, many implementations make it the root of all collections that this user owns, as such a root has to exist anyway if you want to support accessing other users' resources.
* **Calendar/Address book home set** - defined by CalDAV and CardDAV respectively, identifies a collection that contains all calendar/address book collections. Also used for discovery, see e.g. [RFC 4791 section 8.4](https://datatracker.ietf.org/doc/html/rfc4791#section-8.4)
* **Calendar/Address book URL** - identifies a calendar or adress book, essentially a regular WebDAV collection which contains either events or contacts and provides some additional CalDAV/CardDAV-specific attributes. At the moment, `go-webdav` only supports a single calendar/address book. This URL has a relation to the calendar/address book home set: it should be underneath it (or at least equal to it), to be discoverable.

For brevity, I will from now on only focus on CalDAV, but the same applies of course to CardDAV.

**Where we started**

Initially, the server was designed in such a way that everything was served at the root path (`/`). The root path was user principal URL, calendar home set, and calendar URL at the same time. The only other URLs involved were the individual resources (events). This works, but has the following drawbacks:

* Multiple calendars impossible - for multiple calendars, each one must be addressable, requiring URLs like `/calendar1/` etc.
* Serving calendars and address book in the same server impossible - for this to work, a user must be able to discover a calendar home set and an address book home set, which must be distinct from each other, e.g. `/contacts/` and `/calendars/`.
[...]

PR #101, opened 2022-08-24
carddav: switch to one static path layout

bitfehler:
See #100 for details. Obsoletes #99.

### DIFF
--- a/carddav/server.go
+++ b/carddav/server.go
@@ -7,7 +7,9 @@ import (
 	"fmt"
 	"mime"
 	"net/http"
+	"path"
 	"strconv"
+	"strings"
 
 	"github.com/emersion/go-vcard"
 	"github.com/emersion/go-webdav"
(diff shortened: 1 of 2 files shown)


carddav/server.go
```diff
@@ -242,20 +244,28 @@ type backend struct {
...
-	homeSetPath, err := b.Backend.AddressbookHomeSetPath(r.Context())
-	if err != nil {
-		return nil, nil, err
-	}
+const (
+	resourceTypeRoot               resourceType = iota
+	resourceTypeUserPrincipal      resourceType = iota
+	resourceTypeAddressBookHomeSet resourceType = iota
+	resourceTypeAddressBook        resourceType = iota
+	resourceTypeAddressObject      resourceType = iota
```
emersion, line 254: Nit: only the first `resourceType = iota` is necessary, the rest can be omitted.

PR #103, opened 2022-08-31
carddav: evaluate recurrence in match helper 

bitfehler:
The match helper will now properly return recurring events if any of
their recurrences fall into the queried time range. A test for this was
added as well.

This requires bumping the go-ical version the go-webdav depends on, which pulls in github.com/teambition/rrule-go as a dependency.

### DIFF
--- a/caldav/match.go
+++ b/caldav/match.go
@@ -127,9 +127,19 @@ func matchPropFilter(filter PropFilter, comp *ical.Component) (bool, error) {
 func matchCompTimeRange(start, end time.Time, comp *ical.Component) (bool, error) {
 	// See https://datatracker.ietf.org/doc/html/rfc4791#section-9.9
 
-	// TODO handle "infinity" values in query
-	// TODO handle recurring events
+	// evaluate recurring components
+	rset, err := comp.RecurrenceSet(start.Location())
+	if err != nil {
+		return false, err
+	}
+	if rset != nil {
+		// TODO we can only set inclusive to true or false, but really the
+		// start time is inclusive while the end time is not :/
+		return len(rset.Between(start, end, true)) > 0, nil
+	}
 
+	// TODO handle "infinity" values in query
+	// TODO handle more than just events
 	if comp.Name != ical.CompEvent {
 		return false, nil
 	}
--- a/caldav/match_test.go
+++ b/caldav/match_test.go
@@ -253,6 +253,24 @@ END:VCALENDAR`)
 			addrs: []CalendarObject{event1, event2, event3, todo1},
 			want:  []CalendarObject{event1},
 		},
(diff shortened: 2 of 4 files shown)


caldav/match.go
```diff
@@ -127,9 +127,19 @@ func matchPropFilter(filter PropFilter, comp *ical.Component) (bool, error) {
...
-	// TODO handle "infinity" values in query
-	// TODO handle recurring events
+	// evaluate recurring components
+	rset, err := comp.RecurrenceSet(start.Location())
+	if err != nil {
+		return false, err
+	}
+	if rset != nil {
+		// TODO we can only set inclusive to true or false, but really the
+		// start time is inclusive while the end time is not :/
```
emersion, line 137: Note, to fix this, we can just use the iterator API instead of using `Between`: https://github.com/teambition/rrule-go/blob/23177544d08924d34fb19305f949351ade62cbf2/util.go#L145

emersion (approved):
I'm fine landing this as-is, at your discretion.

barkyq:
I have an event (timezone omitted):
```
BEGIN:VCALENDAR
PRODID:DAVx5/4.4.5-ose ical4j/3.2.19 (org.fossify.calendar)
VERSION:2.0
BEGIN:VEVENT
CREATED:20250111T232306Z
DTEND;TZID=Europe/Paris:20250114T200000
DTSTAMP:20250111T235047Z
DTSTART;TZID=Europe/Paris:20250114T190000
RRULE:FREQ=DAILY;COUNT=2;INTERVAL=1
SEQUENCE:5
STATUS:TENTATIVE
SUMMARY:event
UID:FA4733E2-EDE6-454A-BAE1-0AC82E6384AB
X-APPLE-CREATOR-IDENTITY:com.apple.mobilecal
X-APPLE-CREATOR-TEAM-IDENTITY:0000000000
BEGIN:VALARM
ACTION:DISPLAY
DESCRIPTION:marcelo and habib
TRIGGER:-PT30M
END:VALARM
END:VEVENT
```
and a filter:
```
curl -X REPORT http://127.0.0.1/caldav/calendars/default/ -d '<?xml version='1.0' encoding='UTF-8' ?><CAL:calendar-query xmlns="DAV:" xmlns:CAL="urn:ietf:params:xml:ns:caldav"><prop><getetag /></prop><CAL:filter><CAL:comp-filter name="VCALENDAR"><CAL:comp-filter name="VEVENT"><CAL:time-range start="20251001T193931Z" /></CAL:comp-filter></CAL:comp-filter></CAL:filter></CAL:calendar-query>' -H 'content-type: application/xml' -H 'depth: 1'
```
and it does not get returned, but it should...

PR #106, opened 2022-09-15
caldav: fix validation error when VTIMEZONE is after VEVENT

krystiancha:
If in a calendar file VTIMEZONE component is after VEVENT then on the second pass of the `comp` loop `uid` is not blank (set from VEVENT) but `compUID` is blank (set from VTIMEZONE). This wrongly causes error `conflicting UID values in calendar`.

### DIFF
--- a/caldav/caldav.go
+++ b/caldav/caldav.go
@@ -54,7 +54,7 @@ func ValidateCalendarObject(cal *ical.Calendar) (eventType string, uid string, e
 		if uid == "" {
 			uid = compUID
 		}
-		if uid != compUID {
+		if compUID != "" && uid != compUID {
 			return "", "", fmt.Errorf("conflicting UID values in calendar: %s, %s", uid, compUID)
 		}
 	}


caldav/caldav.go
```diff
@@ -54,7 +54,7 @@ func ValidateCalendarObject(cal *ical.Calendar) (eventType string, uid string, e
 		if uid == "" {
 			uid = compUID
 		}
-		if uid != compUID {
+		if comp.Name == ical.CompEvent && uid != compUID {
```
emersion, line 57: Can we check `comp.Name != ical.CompTimezone` instead? The restriction should still be applied to e.g. VTODO, VJOURNAL, and so on.

emersion:
Hm, we should also probably check `compUID != ""`.

krystiancha:
Wouldn't `compUID != "" && uid != compUID` be enough then?

emersion:
Yeah, that should work too.

My comment above was mentioning something else: we should check that VEVENT/VTODO/VJOURNAL/VFREEBUSY have a UID prop. But it's kind of separate.

emersion (approved):
Thanks!

PR #113, opened 2022-10-28
caldav/carddav: return errors instead of panic()

bitfehler:
Removes all remaining panic()'s, so that the server will keep working if clients attempt any of the unsupported features.


emersion:
The panics are inside HTTP handlers, so they won't crash the whole server.

PR #114, opened 2022-10-31
Remove potential quotes around If-Match value

bitfehler:
If the If-Match header is an ETag, the ETag is quoted. Remove the quotes so that the application does not have to deal with this.

Also, add a comment that If-None-Match can potentially be an ETag. This does not seem to be commonly used, though.

### DIFF
--- a/caldav/server.go
+++ b/caldav/server.go
@@ -21,10 +21,10 @@ import (
 type PutCalendarObjectOptions struct {
 	// IfNoneMatch indicates that the client does not want to overwrite
 	// an existing resource.
-	IfNoneMatch bool
+	IfNoneMatch webdav.ConditionalMatch
 	// IfMatch provides the ETag of the resource that the client intends
 	// to overwrite, can be ""
-	IfMatch string
+	IfMatch webdav.ConditionalMatch
 }
 
 // Backend is a CalDAV server backend.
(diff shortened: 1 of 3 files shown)


caldav/server.go
```diff
@@ -499,8 +500,9 @@ func (b *backend) Put(r *http.Request) (*internal.Href, error) {
 	}
 
 	opts := PutCalendarObjectOptions{
+		// TODO according to spec, If-None-Match can also be an ETag
 		IfNoneMatch: r.Header.Get("If-None-Match") == "*",
-		IfMatch:     r.Header.Get("If-Match"),
+		IfMatch:     strings.Trim(r.Header.Get("If-Match"), "\""),
```
emersion, line 505: We already have `internal.ETag` to parse these.

bitfehler: Indeed. Added a helper function for conditional headers that uses it.

internal/elements.go
```diff
@@ -390,6 +390,19 @@ func (etag ETag) String() string {
 	return fmt.Sprintf("%q", string(etag))
 }
 
+// https://datatracker.ietf.org/doc/html/rfc2068#section-14.25 and
+// https://datatracker.ietf.org/doc/html/rfc2068#section-14.26
+func ParseConditionalHeader(hdr string) (string, error) {
+	if hdr == "" || hdr == "*" {
+		return hdr, nil
```
emersion, line 397: The ETag can be `"*"`, in which case this function can't differentiate between that and a wildcard.

Maybe the wildcard checking is best left to the caller, and this function should be called `ParseETag`.

bitfehler: A very good point. Unfortunately, this makes the current API insufficient. As it is external applications (backends) that will have to deal with this, I am proposing to export a type to help them deal with this (see updated change set).

elements.go
```diff
@@ -30,3 +30,24 @@ type groupMembership struct {
 	XMLName xml.Name        `xml:"DAV: group-membership"`
 	Hrefs   []internal.Href `xml:"href"`
 }
+
+// ConditionalHeaderValue represents the value of a conditional header
+// according to RFC 2068 section 14.25 and RFC 2068 section 14.26
+// The (optional) value can either be a wildcard or an ETag.
+type ConditionalHeaderValue string
```
emersion, line 37: Since there are other conditional header fields (e.g. If-Modified-Since), maybe we can find another name? `ETagMatch`? `ConditionalMatch`?

elements.go
```diff
@@ -30,3 +30,24 @@ type groupMembership struct {
...
+
+func (val ConditionalHeaderValue) IsSet() bool {
+	return val != ""
+}
+
+func (val ConditionalHeaderValue) IsWildcard() bool {
+	return val == "*"
+}
+
+func (val ConditionalHeaderValue) ETag() (string, error) {
```
emersion, line 47: Not related to this PR, but the RFC allows multiple ETags in here. Opened https://github.com/emersion/go-webdav/issues/117 to track this.

emersion:
Other than that nitpick, LGTM!

elements.go
bitfehler: Changed to `ConditionalMatch`, thanks!

emersion (approved):
Thanks!

carlaarchuleta707-lgtm:
[``](url)

PR #116, opened 2022-11-15
caldav: use same static path layout as carddav

bitfehler:
This is like #101 but for CalDAV.

There might be some opportunities for shared code, but the devil is in the details. I'd like to wait for support for multiple calendars/address books - it will be fairly easy now, and after that there should be no more big changes. Then I am happy to give it a once-over with regards to code sharing.

Also adds tests similar to the CardDAV ones.


emersion:
Thanks!

PR #122, opened 2023-01-07
Allow PROPFIND request with empty request body

greensea:
A request body is not mandatory in PROPFIND request according to https://www.rfc-editor.org/rfc/rfc4918#section-9.1
If the client does not send a request body, the request should be treated as an AllProp request.


greensea:
I am confused. The ContentLength is a request attribute, not a response header. 

A client may send a request with an empty body, and the http.Request.ContentLength is 0. The patched code check this ContentLength property and make the request treated as an AllProp request.

emersion:
Sorry, I got things mixed up: the client can omit the `Content-Length` header, but still provide a non-empty request body.

emersion:
Superseded by b043bbd9657b

PR #125, opened 2023-07-01
Make sure that `SupportedComponentSet` is properly reflected in `PROPFIND`

cheif:
I started using this project to export tasks over CalDav, more
specifically to `Reminders` on `iOS`/`macOS`. I quickly realized that
even if you specify that `SupportedComponentSet` contains `VTODO`, that
isn't reflected properly wher doing the `PROPFIND`.

This PR should fix that, while keeping the behaviour of defaulting to
`VEVENT` for propfind. Also added some tests to make sure that I didn't
break anything (Which I hope I didn't :sweat_smile:).


cheif:
🙏

PR #126, opened 2023-07-05
Make sure to handle a PROPFIND to `/`

cheif:
It seems like the Reminders app in `iOS`/`macOS` does this request as the first thing when setting up an account, so it seems reasonable to handle it for us.

This just returns the most basic current-user-principal now, but that should hopefully be enough to continue the process.

### DIFF
--- a/caldav/server.go
+++ b/caldav/server.go
@@ -372,11 +372,7 @@ func (b *backend) PropFind(r *http.Request, propfind *internal.PropFind, depth i
 
 	switch resType {
 	case resourceTypeRoot:
-		_, err := b.Backend.CurrentUserPrincipal(r.Context())
-		if err != nil {
-			return nil, err
-		}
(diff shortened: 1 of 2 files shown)


emersion:
Hm, actually, no, this is the root, not the user principal… Nevermind about using `webdav.ServePrincipal`.

cheif:
Yeah, I'm not competent enough in `*Dav` to know what the most reasonable approach is here. 

I just assume that it's proper to respond to the `<current-user-principal>` request no matter what the path is.

emersion:
Okay. Can we copy-paste `propFindUserPrincipal` over to a new `propFindRoot` function which only returns the current user principal prop?

Also I think the `CurrentUserPrincipal` call can be dropped, its result is unused.

cheif:
@emersion Done! 😄

emersion (approved):
LGTM, thanks!

PR #127, opened 2023-07-13
Multi-calendar support

cheif:
This is my stab at adding multi-calendar support for the CalDav implementation, I have added some questions/thoughts in the PR which I'd love to get some feedback on before I finish it up.

# TODO
- [ ] Write some comments about how to use the multi-calendar support

### DIFF
--- a/caldav/server.go
+++ b/caldav/server.go
@@ -31,6 +31,7 @@ type PutCalendarObjectOptions struct {
 type Backend interface {
 	CalendarHomeSetPath(ctx context.Context) (string, error)
 	ListCalendars(ctx context.Context) ([]Calendar, error)
+	GetCalendar(ctx context.Context, path string) (*Calendar, error)
 	GetCalendarObject(ctx context.Context, path string, req *CalendarCompRequest) (*CalendarObject, error)
 	ListCalendarObjects(ctx context.Context, path string, req *CalendarCompRequest) ([]CalendarObject, error)
 	QueryCalendarObjects(ctx context.Context, query *CalendarQuery) ([]CalendarObject, error)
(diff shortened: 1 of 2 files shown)


caldav/server.go
```diff
@@ -42,6 +40,11 @@ type Backend interface {
 	webdav.UserPrincipalBackend
 }
 
+type MultiCalendarBackend interface {
```
cheif, line 43: I'm not familiar enough with `go` to know if this is a reasonable way to implement this change. This means we're 100% backwards compatible, and it's up for the user of the library to conform to `MulitCalendarBackend` to get that support, I think the biggest drawback to this is that discoverability suffers, and it adds some more possible paths for the code to take here. Maybe it would be OK to break the API instead, adopting the changes should be very straightforward (just return a `[]*Calendar` instead of `*Calendar`) etc.

caldav/server.go
```diff
@@ -42,6 +40,11 @@ type Backend interface {
...
+	Calendars(ctx context.Context) ([]*Calendar, error)
```
cheif, line 44: I'm also not sure if it makes sense to use `[]*Calendar` here, or if we could just go with `[]Calendar`? I went with this now, since it was already used in `propFindAllCalendars`

caldav/server.go
```diff
@@ -42,6 +40,11 @@ type Backend interface {
...
+	ListCalendarObjectsForCalendar(ctx context.Context, cal *Calendar, req *CalendarCompRequest) ([]CalendarObject, error)
```
cheif, line 45: I went with this take a `*Calendar` now, but in some cases similar methods take a `path string` instead, I'm not sure which one we should go for?

cheif:
@emersion Any thoughts on this?

cheif:
Now I've addressed all comments (I hope), so please take a new look when there's time 😄

caldav/server.go
```diff
@@ -32,9 +30,9 @@ type PutCalendarObjectOptions struct {
 // Backend is a CalDAV server backend.
 type Backend interface {
 	CalendarHomeSetPath(ctx context.Context) (string, error)
-	Calendar(ctx context.Context) (*Calendar, error)
+	Calendars(ctx context.Context) ([]Calendar, error)
```
emersion, line 33: For symmetry with calendar objects, maybe we could rename this to `ListCalendars`?

emersion:
Just two nits, overall LGTM!

emersion (approved):
Thank you!

PR #129, opened 2023-09-03
fix: check etag quotes before unquote

in0rdr:
I encountered the same issue with non-standard (unquoted) etags from my caldav server as my predecessors:
* https://github.com/emersion/go-webdav/issues/50
* https://github.com/emersion/go-webdav/pull/69

```bash
go test
--- FAIL: TestETagUnmarshalText (0.00s)
    elements_test.go:79: UnmarshalText failed for ETag: webdav: failed to unquote ETag: invalid syntax
FAIL
exit status 1
FAIL    github.com/emersion/go-webdav/internal  0.006s
```

This patch will check the etag first, before calling the [`strconv.Unquote`](https://pkg.go.dev/strconv#Unquote) function. A string without quotes will not be unquoted.

### DIFF
--- a/internal/elements.go
+++ b/internal/elements.go
@@ -374,11 +374,16 @@ type GetETag struct {
 type ETag string
 
 func (etag *ETag) UnmarshalText(b []byte) error {
-	s, err := strconv.Unquote(string(b))
-	if err != nil {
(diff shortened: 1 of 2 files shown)


emersion:
This will not behave properly when encoding the value.

PR #132, opened 2023-10-07
Patch to not fail when eTag not present

jtagcat:
Feel free to close if this is not RFC-compliant. The server I am interacting with doesn't provide an eTag.


jtagcat:
Superseded by https://github.com/emersion/go-webdav/pull/174

PR #134, opened 2023-12-15
caldav: properly defer-close body of caldav put request

sbinet:


emersion:
Does this is a real bug you're seeing?

`populateCalendarObject` only looks at header fields. I don't think the response body is supposed to contain anything.

The change is not harmful or anything, I just think it doesn't make a difference.

sbinet:
I've found out about this mistake while trying to understand why an `*ical.Calendar` I was uploading to my framagenda was coming back as a `caldav.CalendarObject` with a `0` `ModTime` value.

adding the `defer` didn't fix my thing.
(but I'd say it's less surprising w/ the `defer`)

emersion:
I've changed the function to take the header as input instead of the response: 0e58dbb00392fa03ed2958d5ab3fdf31d664b75b. This should make the code less surprising.

sbinet:
nice.

PR #135, opened 2023-12-15
internal: fix always-true interface comparison

sbinet:
This CL corrects the following bug uncovered by staticcheck:

```
  internal/elements.go:148:6: this comparison is always true (SA4023)
    internal/elements.go:146:18: the lhs of the comparison gets its value from here and has a concrete type
```

### DIFF
--- a/internal/elements.go
+++ b/internal/elements.go
@@ -143,7 +143,10 @@ func (resp *Response) Err() error {
 		return nil
 	}
 
-	var err error = resp.Error
+	var err error
+	if resp.Error != nil {
+		err = resp.Error
+	}
 	if resp.ResponseDescription != "" {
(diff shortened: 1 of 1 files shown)


emersion:
Ouch. That's the usual "nil pointer but not actually nil interface" footgun right?

Can we ensure that we never return a nil `*Error` (which ends up being a non-nil `error`)? Even if `resp.ResponseDescription == ""`?

Probably this would help:

```
var err error
if resp.Error != nil {
    err = resp.Error
}
```

sbinet:
> That's the usual "nil pointer but not actually nil interface" footgun right?

yep, that's the one.

PTAL.

emersion (approved):
Thank you!

PR #136, opened 2023-12-15
webdav: introduce CopyOptions

emersion:
Not super happy about the negative bool fields…


emersion:
Another solution would be to use pointers:

```go
type CopyOptions struct {
    Recursive *bool // defaults to true
    Overwrite *bool // defaults to true
}
```

But it's cumbersome to set the fields (requires an intermediate variable).

PR #138, opened 2024-01-13
Adding encrypted client

aodhan-domhnaill:
*This is a draft and not done. I'm just checking if this is interesting.*

I personally use WebDav using e2e encryption for storing data on my personal server. I use rclone for this, but rclone is lacking since it only supports symmetric encryption. With symmetric only, it couldn't be extended to support e2e Caldav or e2e Carddav where someone sends e2e calendar events to you.

This is a draft to show how I was intending to add e2ee support. If there is interest, I will finish the PR.


emersion:
Thanks for the patch, but I'd rather not add a client for an encryption scheme not defined in a WebDAV-related RFC. I'd suggest to keep this in a separate repository instead :)

aodhan-domhnaill:
Fair enough

PR #139, opened 2024-02-01
caldav: fix match on open time ranges

bitfehler:
Matches on open time ranges (i.e. no end date) were not properly handled, as `end` is simply the zero time, which confuses the `.Before()` and `.After()` logic employed here.

This commit fixes that by adding the appropriate `.IsZero()` checks and also adds a test case.

The current behavior unfortunately broke compatibility with DAVx5, which by default queries only events less than 90 days ago (by using an open time range).

### DIFF
--- a/caldav/match.go
+++ b/caldav/match.go
@@ -138,7 +138,6 @@ func matchCompTimeRange(start, end time.Time, comp *ical.Component) (bool, error
 		return len(rset.Between(start, end, true)) > 0, nil
 	}
 
-	// TODO handle "infinity" values in query
 	// TODO handle more than just events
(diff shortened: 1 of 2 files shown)


emersion:
Should we do the same in `matchPropTimeRange`?

bitfehler:
Good point, added there as well, and removed the TODOs about infinity, because this is exactly that

emersion:
Also it seems like `matchPropFilter` ignores `filter.TextMatch` when `Start` is defined, not sure this is the correct behavior? But let's leave this for another patch.

emersion (approved):
Thanks!

PR #140, opened 2024-02-02
caldav: add path to interface QueryCalendarObjects

bitfehler:
This was missing for proper multi-calendar support.

--

Not sure why this wasn't added in #127, but it looks like just oversight to me? A query (a.k.a. a REPORT request) has to be made against a specific calendar (or a calendar object resource), so the backend has to know which one.

PR #141, opened 2024-02-02
carddav: support multiple address books

bitfehler:
This is the equivalent of #127 (and #140) for CardDAV and finally allows backends to serve different address books to different users.

While I'm breaking the interface, correct one last instance of "Addressbook" to "AddressBook" (in `AddressBookHomeSetPath`).


bitfehler:
I forgot something, will fix

bitfehler:
Had actually just forgotten to remove the TODO comments :innocent: sorry  about the noise

emersion (approved):
Thanks!

PR #143, opened 2024-02-07
carddav: support address book creation/deletion

bitfehler:
Now that the handling for multiple address books is in place, this commit adds initial support for creation and deletion of address books.

These operations obviously require support from the backend, so the interface gains two new methods. All properties of the address book passed to `CreateAddressBook()` may be unset, except for the path (e.g. when a client sends a MKCOL request without a body. It is up to the backend to put any desired default values in place. As is, the parsing of the request body still needs some improvements, but correctly handles the most common properties.

Using this as is (in tokidoki), I can successfully create and delete address books using DavX5. The validation is not great yet, but since this touches the interface again, I wanted to make sure early on that this looks like the right approach to you.

### DIFF
--- a/carddav/elements.go
+++ b/carddav/elements.go
@@ -211,3 +211,11 @@ func (r *reportReq) UnmarshalXML(d *xml.Decoder, start xml.StartElement) error {
 
 	return d.DecodeElement(v, &start)
 }
+
+type mkcolReq struct {
+	XMLName      xml.Name               `xml:"DAV: mkcol"`
+	ResourceType internal.ResourceType  `xml:"set>prop>resourcetype"`
+	DisplayName  string                 `xml:"set>prop>displayname"`
+	Description  addressbookDescription `xml:"set>prop>addressbook-description"`
+	// TODO this could theoretically contain all addressbook properties?
+}
--- a/carddav/server.go
+++ b/carddav/server.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/xml"
 	"fmt"
+	"io"
 	"mime"
 	"net/http"
 	"path"
@@ -30,6 +31,8 @@ type Backend interface {
 	AddressBookHomeSetPath(ctx context.Context) (string, error)
 	ListAddressBooks(ctx context.Context) ([]AddressBook, error)
 	GetAddressBook(ctx context.Context, path string) (*AddressBook, error)
+	CreateAddressBook(ctx context.Context, addressBook AddressBook) error
+	DeleteAddressBook(ctx context.Context, path string) error
 	GetAddressObject(ctx context.Context, path string, req *AddressDataRequest) (*AddressObject, error)
 	ListAddressObjects(ctx context.Context, path string, req *AddressDataRequest) ([]AddressObject, error)
 	QueryAddressObjects(ctx context.Context, path string, query *AddressBookQuery) ([]AddressObject, error)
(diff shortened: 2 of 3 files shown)


emersion:
Yeah, in general this looks like a good approach to me!

carddav/server.go
```diff
@@ -682,11 +684,42 @@ func (b *backend) Put(r *http.Request) (*internal.Href, error) {
...
+		return internal.HTTPErrorf(http.StatusForbidden, "carddav: address book creation not allowed at given location")
+	}
+
+	ab := AddressBook{
+		Path: r.URL.Path,
+	}
+
+	// Check if a request body was sent
+	var buf []byte
+	_, err := r.Body.Read(buf)
```
emersion, line 707: Hm, is this a reliable way to check for empty request body?

It seems like it might be:

> If len(p) == 0, Read should always return n == 0. It may return a non-nil error if some error condition is known, such as EOF. 

Nit: it would be simpler to `Read(nil)`.

Ref https://github.com/emersion/go-webdav/commit/b043bbd9657b2c47250648db9889d78295f7da27

carddav/server.go
```diff
@@ -682,11 +684,42 @@ func (b *backend) Put(r *http.Request) (*internal.Href, error) {
...
+	if err == nil {
```
emersion, line 708: We should probably return the error if it's non-nil and non-EOF.

bitfehler:
Cool, thanks. I added another commit (will squash at the end). It does two things, both of which look slightly ugly to me, so again I'd appreciate your opinion!

1. The parsing of the `<mkcol>` element now checks that the `<resourcetype>` element contains exactly `<cDAV: ollection/>` and `<CAL: calendar>`, but it's ugly AF. Is there any better way to do it? I couldn't find one...
2. You are already adding comments about the empty body thing, so i'll just reply there instead :slightly_smiling_face:

carddav/server.go
bitfehler: > Hm, is this a reliable way to check for empty request body?
> 
> It seems like it might be:
> 
> > If len(p) == 0, Read should always return n == 0. It may return a non-nil error if some error condition is known, such as EOF.
> 
> Nit: it would be simpler to `Read(nil)`.

So, I am hoping it is. If there is anything better, I am certainly open.

Also, it is up for discussion if we want to support this at all. A MKCOL with an empty body is valid according to [WebDAV](https://datatracker.ietf.org/doc/html/rfc4918#section-9.3), but I am honestly not sure if any CardDAV/CalDAV clients actually do it. Still, I thought it might be nice to at least not prevent it from working.

carddav/elements.go
```diff
@@ -211,3 +211,33 @@ func (r *reportReq) UnmarshalXML(d *xml.Decoder, start xml.StartElement) error {
 
 	return d.DecodeElement(v, &start)
 }
+
+type rtdef struct {
+	IsCollection  bool
+	IsAddressBook bool
+}
+
+type resourcetype struct {
```
emersion, line 220: We already have `internal.ResourceType`, maybe it's good enough for our purposes?

bitfehler: Thanks for reminding me of that! I updated the second commit, it looks much better now

carddav/server.go
bitfehler: yep

bitfehler:
I think this is a good start now. Let me know what you think.

emersion:
FYI, I've made a small adjustment: https://github.com/emersion/go-webdav/commit/fbcd08d64ad28a3f902d9409551237af6f4decf7

PR #145, opened 2024-02-08
*: displayname, desription are optional

bitfehler:
Both the displayname and the description can be absent for both calendars and address books. If this is the case they should not show up in PROPFIND responses as empty string.


emersion:
Good catch, thanks!

PR #146, opened 2024-02-08
internal: no status element in propstat responses

bitfehler:
Responses that contain propstat elements do not contain their own top-level status element, only the status elements inside the propstat element.

See https://datatracker.ietf.org/doc/html/rfc4918#section-14.24 or any of the examples for PROPFIND/PROPPATCH, starting e.g. here: https://datatracker.ietf.org/doc/html/rfc4918#section-9.1.3


emersion (approved):
Indeed, thanks!

PR #147, opened 2024-02-08
carddav: PROPPATCH support for address books

bitfehler:
The groundwork for address objects is also there, but it's not fully implemented.

I'll be honest: none of the usual suspects (DavX5, TB/CardBook, Evolution) actually use this, but that just makes me mad. They all rename address books client side only (or not at all). However, since PROPPATCH is a standard WebDAV operation, I successfully renamed my address books on the server using cadaver. I still think this is worth implementing, because if any sane client ever comes along, they might really want to use this.

Regarding the support for address objects, it might be worth taking a closer look at the worthwhileness before going through with it. I suspect that again no major client actually uses it, and for address objects it's also slightly less useful (because e.g. the "displayname" property is likely not used, because a contact is displayed by its name, etc.).


emersion:
Does it help clients to advertise PROPPATCH here? https://github.com/emersion/go-webdav/blob/25f1014ef25f32967c8e17da16fba6f5c78886c0/carddav/server.go#L281

carddav/server.go
```diff
@@ -614,43 +615,160 @@ func (b *backend) propFindAllAddressObjects(ctx context.Context, propfind *inter
...
-	}
-
+	resType := b.resourceTypeAtPath(r.URL.Path)
 	resp := internal.NewOKResponse(r.URL.Path)
 
-	if r.URL.Path == homeSetPath {
-		// TODO: support PROPPATCH for address books
+	switch resType {
+	case resourceTypeAddressBook:
+		ab, err := b.Backend.GetAddressBook(r.Context(), r.URL.Path)
```
emersion, line 623: Hm, this is not atomic… If the address book is updated by another client after `GetAddressBook`, but before `UpdateAddressBook`, then we'll be passing stale data.

Can we instead indicate somehow which fields are being updated? This could be done by passing a different struct with pointers (nil would mean "leave as is"), or by passing a bitmask of updated fields, for instance.

carddav/server.go
```diff
@@ -614,43 +615,160 @@ func (b *backend) propFindAllAddressObjects(ctx context.Context, propfind *inter
...
+				rxn, ok := raw.XMLName()
+				if !ok {
+					return nil, fmt.Errorf("failed to parse properties")
+				}
+				emptyVal := internal.NewRawXMLElement(rxn, nil, nil)
+				if err := resp.EncodeProp(http.StatusMethodNotAllowed, emptyVal); err != nil {
+					return nil, err
+				}
 			}
 		}
```
emersion, line 675: Do we have to write out a multi-status here? Can't we just reply with a toplevel HTTP status instead? Something like `internal.HTTPErrorf`?

carddav/server.go
```diff
@@ -614,43 +615,160 @@ func (b *backend) propFindAllAddressObjects(ctx context.Context, propfind *inter
...
+					return err
+				}
+			case addressBookDescriptionName:
+				ab.Description = ""
+				if err := resp.EncodeProp(http.StatusOK, desc); err != nil {
+					return err
+				}
+			default:
+				emptyVal := internal.NewRawXMLElement(rxn, nil, nil)
+				if err := resp.EncodeProp(http.StatusNotImplemented, emptyVal); err != nil {
```
emersion, line 705: Would a 404 be more appropriate for the properties we don't know about?

bitfehler:
Changed the interface so that the backend can implement UpdateAddressbook atomically, also simplified some things a bit.

PR #151, opened 2024-04-09
webdav: change FileSystem.Create to give implementations more control

DeepDiver1975:
The existing interface seems to be driven be the local file system implementation.

For e.g. an S3 implementation of PUT a bit more control over is required. 
By passing in request.Body into FileSystem.PutFile() the implementation can directly stream the request content to the target or perform chunked/multi-part upload or what ever is necessary ...

### DIFF
--- a/fs_local.go
+++ b/fs_local.go
@@ -114,13 +114,25 @@ func (fs LocalFileSystem) ReadDir(ctx context.Context, name string, recursive bo
 	return l, errFromOS(err)
 }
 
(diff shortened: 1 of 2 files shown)


DeepDiver1975:
The PUT method can also be used to overwrite an existing resource. `FileSystem.Create()` feels a bit wrong.

emersion:
`os.Create` behaves the same way though: it overwrites any previous file.

emersion (approved):
Thanks!

PR #153, opened 2024-04-09
webdav: add size to FileSystem.Create()

DeepDiver1975:
In some cases the total size is necessary upfront to perform the upload to an external system.

And even if not it is good to have it as a minimal test to see if data was transmitted completely ....

Open question:
how to address other valuable headers like checksums or what ever ....

maybe we pass the full request object to Create()?
Or body + headers?

@emersion any objections/idea?

### DIFF
--- a/fs_local.go
+++ b/fs_local.go
@@ -114,7 +114,7 @@ func (fs LocalFileSystem) ReadDir(ctx context.Context, name string, recursive bo
 	return l, errFromOS(err)
 }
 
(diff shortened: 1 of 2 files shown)


emersion:
Hm right… I think the use-case is reasonable, but backends should keep in mind that the size is just a hint, it may be unset (-1). Also, HTTP compression (Content-Encoding) will always leave Content-Length unset.

PR #154, opened 2024-04-09
webdav: add http.Header to FileSystem.Create()

DeepDiver1975:
alternative to #153


emersion:
I'm not a huge fan of exposing generic HTTP stuff to backends in general.

If there is a custom HTTP header that a library user wants to pass to the backend, they can use the context.

PR #155, opened 2024-04-11
internal: use application/xml instead of text/xml which is deprecated

DeepDiver1975:
see https://datatracker.ietf.org/doc/html/rfc4918#section-8.2


emersion:
Thanks!

PR #158, opened 2024-04-17
webdav: FileSystem.Create() returns FileInfo and is used to set PUT r…

DeepDiver1975:
…esponse headers

PR #159, opened 2024-04-17
webdav: respond PUT request with 204/No Content in case the file already existed before putting

DeepDiver1975:
to be merged after #158

### DIFF
--- a/fs_local.go
+++ b/fs_local.go
@@ -114,25 +114,36 @@ func (fs LocalFileSystem) ReadDir(ctx context.Context, name string, recursive bo
 	return l, errFromOS(err)
 }
 
-func (fs LocalFileSystem) Create(ctx context.Context, name string, body io.ReadCloser) (*FileInfo, error) {
+func (fs LocalFileSystem) Create(ctx context.Context, name string, body io.ReadCloser) (*FileInfo, bool, error) {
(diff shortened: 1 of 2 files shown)


DeepDiver1975:
@emersion this one is also ready now. THX

fs_local.go
```diff
@@ -114,25 +114,36 @@ func (fs LocalFileSystem) ReadDir(ctx context.Context, name string, recursive bo
...
 
-func (fs LocalFileSystem) Create(ctx context.Context, name string, body io.ReadCloser) (*FileInfo, error) {
+func (fs LocalFileSystem) Create(ctx context.Context, name string, body io.ReadCloser) (*FileInfo, bool, error) {
 	p, err := fs.localPath(name)
 	if err != nil {
-		return nil, err
+		return nil, false, err
+	}
+	created := false
+	fi, _ := fs.Stat(ctx, name)
```
emersion, line 123: I think it should be `p` instead of `name`.

Oops, it sounds like I didn't catch the issue when reviewing the previous PR: the `Stat` call at the end of the function has the same issue.

Nit: could be simplified with something like:

```go
created := fi == nil
``

emersion: Oh yes, indeed!

emersion:
Thanks!

PR #161, opened 2024-04-17
internal: add PropFindValue

emersion:
NewPropFindResponse uses callbacks to lazily build the response. However, some props are static: they don't require any processing to generate. Add a small helper to reduce boilerplate a bit.


oliverpool (approved):
This change looks like a net improvement!

It would be nice to merge it, to prevent conflicts for future merge requests (I was almost ready to also make a fix for thunderbid, like #170 but with tests :)

barkyq:
looks cleaner

barkyq:
:fireworks:

PR #162, opened 2024-04-19
Update go-ical

bitfehler:
It's only a dependency update in go-ical, but it allows gettin rid of rrule-go v.1.7, which is nice.


emersion (approved):
Thanks!

PR #164, opened 2024-06-06
webdav: PUT response has no body and therefore should not have a cont…

DeepDiver1975:
…ent length header


emersion (approved):
Indeed, [RFC 9110 section 8.6](https://www.rfc-editor.org/rfc/rfc9110#section-8.6) says that a `Content-Length` header is invalid with a 204.

Thanks!

PR #105, opened 2022-08-31
caldav: support more features for recurring events

bitfehler:
The protocol allows clients to ask the server to e.g. expand recurring
events to a set of simple events in a given time-frame, so that the
client does not have to calculate the recurrences.

This commit adds the foundation for support for expand [1],
limit-recurrence-set [2], and limit-freebusy-set [3] functionality.
However, the actual transformation of events returned to clients is not
yet implemented.

[1]: https://datatracker.ietf.org/doc/html/rfc4791#section-9.6.5
[2]: https://datatracker.ietf.org/doc/html/rfc4791#section-9.6.6
[3]: https://datatracker.ietf.org/doc/html/rfc4791#section-9.6.7

PR #167, opened 2024-08-18
Added expand request

jonyTF:
I needed the expand request for a project I'm working on (wanted recurring events to show at the correct date), so I went ahead and added it!

### DIFF
--- a/.build.yml
+++ b/.build.yml
@@ -2,7 +2,7 @@ image: alpine/edge
 packages:
   - go
 sources:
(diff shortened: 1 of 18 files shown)


emersion:
The PR contains all of the import renames.

emersion:
And it sounds like this is just a subset of #105?

jonyTF:
oops sorry about that, fixed it in this updated PR: #168

Also, yes it's just a subset of #105 but it seems that guy hasn't worked on that PR in two years

PR #168, opened 2024-08-21
caldav: add expand request to client

jonyTF:
I needed the expand request for a project I'm working on (wanted recurring events to show at the correct date), so I went ahead and added it!

This is a pull request fixing my previous pull request that had import renames polluting the commit history


emersion (approved):
Thanks!

PR #170, opened 2024-09-23
implements CurrentUserPrivilegeSet to add compatibility with thunderb…

mheers:
…ird address book

### DIFF
--- a/internal/elements.go
+++ b/internal/elements.go
@@ -21,7 +21,8 @@ var (
 	GetLastModifiedName  = xml.Name{Namespace, "getlastmodified"}
 	GetETagName          = xml.Name{Namespace, "getetag"}
 
-	CurrentUserPrincipalName = xml.Name{Namespace, "current-user-principal"}
+	CurrentUserPrincipalName    = xml.Name{Namespace, "current-user-principal"}
+	CurrentUserPrivilegeSetName = xml.Name{Namespace, "current-user-privilege-set"}
 )
 
 type Status struct {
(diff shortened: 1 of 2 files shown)


internal/elements.go
```diff
@@ -417,6 +418,16 @@ type CurrentUserPrincipal struct {
...
 
+// https://datatracker.ietf.org/doc/html/rfc3744#section-5.4
+type CurrentUserPrivilegeSet struct {
+	XMLName    xml.Name    `xml:"DAV: current-user-principal"`
+	Privileges []Privilege `xml:"privilege"`
+}
+type Privilege struct {
+	Read  bool `xml:"read"`
+	Write bool `xml:"write"`
+}
```
oliverpool, line 429: In my case, I got it to work with Thunderbird locally with the following structure (didn't tried yours TBH):
```suggestion
	Privileges []Privilege `xml:"privilege>dynamic"`
}
type Privilege struct {
	XMLName xml.Name
}
```

And then return `read`, `write` like this:
```go
return &internal.CurrentUserPrivilegeSet{Privileges: []internal.Privilege{{XMLName: xml.Name{"", "read"}}, {XMLName: xml.Name{"", "write"}}}}, nil
```

(ideally the read/write privileges information should come from the backend)

oliverpool (commented):
I just got the exact same issue myself and solved it before seeing your PR :see_no_evil:

oliverpool:
Reference: https://stackoverflow.com/a/57124617/3207406

internal/elements.go
```diff
@@ -417,6 +418,15 @@ type CurrentUserPrincipal struct {
 	Unauthenticated *struct{} `xml:"unauthenticated,omitempty"`
 }
 
+// https://datatracker.ietf.org/doc/html/rfc3744#section-5.4
+type CurrentUserPrivilegeSet struct {
+	XMLName    xml.Name    `xml:"DAV: current-user-privilege-set"`
+	Privileges []Privilege `xml:"privilege>dynamic"`
```
emersion, line 424: Nit: this file use the same name as the XML element even if there may be multiple.

emersion, line 424: Why `dynamic`? This doesn't appear in the RFC AFAIK.

internal/elements.go
```diff
@@ -417,6 +418,15 @@ type CurrentUserPrincipal struct {
...
+}
+type Privilege struct {
```
emersion, line 426: Instead of a list, could we use a struct with fields for read/write?

internal/elements.go
emersion: Seems like the second part of the suggestion hasn't been applied. (But I prefer the struct approach.)

emersion:
Superseded by a16253aa97a7f8cfc2d89be6dfa7b36719b45551

PR #171, opened 2024-10-05
caldav, carddav: redirect on wrong `principalPath` or `homeSetPath` instead of empty MultiStatus response

oliverpool:
Before this PR if the client was requesting a wrong `principalPath` or `homeSetPath` it would get an empty `207 Multistatus` response.

With this PR, the client would get a 308 redirect response instead.

Concrete usecase: when an iOS user makes a change to its username, iOS will automatically update the principal URL if it receives a redirect response.

I tried to send the redirection inside the 207 (using `location`), but couldn't get it to work (don't know if I did something wrong, or if iOS does not support it).

_PS: this change is available in the `main` branch of my fork: https://github.com/oliverpool/go-webdav/_

### DIFF
--- a/caldav/server.go
+++ b/caldav/server.go
@@ -366,7 +366,7 @@ func (b *backend) HeadGet(w http.ResponseWriter, r *http.Request) error {
 	return nil
 }
 
-func (b *backend) PropFind(r *http.Request, propfind *internal.PropFind, depth internal.Depth) (*internal.MultiStatus, error) {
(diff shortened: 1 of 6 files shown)


emersion:
This doesn't make sense to me. A wrong username generally just means that it's wrong, not that it's an old one.

In general I wouldn't recommend using identifiers which can change in a URL. An immutable user ID alleviates the issue.

oliverpool:
You are right, this is an obviously wrong solution.

Maybe I find a better way

PR #172, opened 2024-10-06
Initial ACL implementation

oliverpool:
Also aiming to support thunderbird (like #170), I took a deeper look at the RFC7344.

This is my current state (still very WIP), but I thought that sharing it earlier would be good.

The static `CurrentUserPrivilegeSet` of @mheers PR would look like this with this PR:
```go
func (b *backend) propFindAddressObject(ctx context.Context, propfind *internal.PropFind, ao *AddressObject) (*internal.Response, error) {
	props := map[xml.Name]internal.PropFindFunc{
		internal.CurrentUserPrivilegeSetName: func(*internal.RawXMLValue) (interface{}, error) {
			return &internal.CurrentUserPrivilegeSet{
				Privileges: []internal.Privilege{
					internal.NewPrivilege(internal.Read),
					internal.NewPrivilege(internal.Write),
				},
			}, nil
		},
```

Ideally, I would like to expose some knobs to the Backend, to tweak the actually returned privileges.

@emersion I would love to have your early feedback, if you think that this is going in a sensible direction!

I did quite a lot of copy/paste from the RFC to ease the understanding of the code (with the dream of having something like https://lobste.rs/s/ogdfxh/cross_referencing_code_specs_for at some point in the future: RFC coverage / cross-referencing)

### DIFF
--- a/internal/acl.go
+++ b/internal/acl.go
@@ -1,0 +1,538 @@
+package internal
+
+import (
+	"encoding/xml"
+)
+
+func NewPrivilege(name xml.Name) Privilege {
+	return Privilege{
+		Raw: NewRawXMLElement(name, nil, nil),
+	}
+}
+
+type Privilege struct {
+	XMLName xml.Name     `xml:"DAV: privilege"`
+	Raw     *RawXMLValue `xml:",any"`
+}
+
+func (p Privilege) Is(target xml.Name) bool {
+	got, ok := p.Raw.XMLName()
+	return ok && got == target
+}
+
+var (
+	/*
+	   rfc3744#section-3.1
+
+	   The read privilege controls methods that return information about the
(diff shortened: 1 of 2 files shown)


internal/acl.go
```diff
@@ -0,0 +1,548 @@
...
+// convenience CurrentUserPrivilegeSet
+var (
+	CurrentUserPrivilegeSetReadOnly = CurrentUserPrivilegeSet{
+		Privilege: []Privilege{NewPrivilege(Read)},
+	}
+	CurrentUserPrivilegeSetReadWrite = CurrentUserPrivilegeSet{
+		Privilege: []Privilege{NewPrivilege(Read), NewPrivilege(Write)},
+	}
+)
+
```
oliverpool, line 363: Well hidden in this huge file :wink:

Issue #165, opened 2024-07-03
Feature Request: Support for out-of-spec unquoted getetag values


EtienneBruines:
Some WebDAV servers do not return quoted etag values (to make them re-usable for the ETag HTTP response header), e.g. https://github.com/haiwen/seafdav/issues/82 and https://github.com/photoprism/photoprism/issues/3495 even though they should.

Would it be possible/reasonable to accept unquoted ETag values? To make this library compatible with a wider range of servers?

Potentially like https://github.com/emersion/go-webdav/pull/69 although I was thinking modifying https://github.com/emersion/go-webdav/blob/810c51fa2d38ba62ca3971524f9cc322a6a504f4/internal/elements.go#L382 to not return an error if it's simply an alphanumerical string (a hash or something) that just doesn't have quotes. 

What do you think?

PR #174, opened 2024-10-26
internal: fixed ETag UnmarshalText problem for unquoted text

borisovdev:
For some caldav providers, it wasn't possible to unmarshal unquoted text for etag

Referenced to #165, #69


emersion:
This is non-standard, so not sure I want to maintain this.

This also unnecessarily uses regular expressions, and falls apart when marshaling etags.

borisovdev:
> This is non-standard, so not sure I want to maintain this.
> 
> This also unnecessarily uses regular expressions, and falls apart when marshaling etags.

Thanks for the quick reply and your time!
Yes, I have already familiarized myself with other similar issues. The reality is that many providers do not strictly follow the RFC in this place. Should we send people to live in a fork just because the rules say so? Often we do, but it seems not in this case. Please give it some thought:)

Thanks for pointing out MarshalText, I should be more careful. I'm marking PR as draft and will finalize it as soon as I have free time.

Translated with DeepL.com (free version)

borisovdev:
I like to clarify details, to avoid misunderstandings) We are talking about https://datatracker.ietf.org/doc/html/rfc2616#section-3.11, right?

About the review:
> This also unnecessarily uses regular expressions and falls apart when marshaling etags.

1. removed regexp from implementation
2. I thought of possible problems with marshaling in my head, but didn't find them. In the extended implementation of UnmarshalText an unquoted string is returned anyway. I added new tests and modified the old one

> This is non-standard, so not sure I want to maintain this.

the fate of this item is in your hands:)

Rebased and updated the branch

barkyq:
another option would be to check if the `b` byte slice has " and " as its first and last bytes, and, if not, add them and proceed as before.

barkyq:
another comment, if this is used (by changing `internal`), it also applies to server side, which seems like it would make the servers accept the non-standard behaviour of unquoted etags.

berzdev:
This Pull requests is great! I now can use CardDAV from Mailbox.org!

Please merge it!

emersion:
This PR treats weak ETags as strong ETags, and breaks when marshaling an ETag value.

Issue #117, opened 2022-11-22
Add support for multiple ETags in If-Match/If-None-Match


emersion:
e.g.

     If-Match: "xyzzy", "r2d2xxxx", "c3piozzzz"

Ref https://www.rfc-editor.org/rfc/rfc7232

PR #180, opened 2025-01-13
webdav: ConditionalMatch with multiple etags

barkyq:
#117


webdav.go
```diff
@@ -54,10 +55,36 @@ func (val ConditionalMatch) ETag() (string, error) {
...
+// Returns:
+//   - isSet: Indicates if the ConditionalMatch has any ETags set, or is wildcard.
+//   - match: True if the etag matches any of the ETags in ConditionalMatch, false otherwise.
+//   - err: An error if there was a problem parsing one of the ETags.
+//     If an error occurs during parsing, match will be set to false, but isSet will be true.
+//     Callers should check for a non-nil error to ensure the match result is valid.
+//
+// The function returns early if no ETags are set (isSet is false) or if a wildcard (*) is used,
+// in which case all ETags match. For multiple ETags, it checks each one until a match is found or all are checked.
+func (val ConditionalMatch) MatchETag(etag string) (isSet bool, match bool, err error) {
```
barkyq, line 72: rationale for returning `isSet`

the MatchETag function should check whether val.IsSet() is non-nil, otherwise it leads to the wrong behaviour.

this seemed the only reasonable solution (beyond implementors inspecting error types), and reduces boilerplate code for implementations.

PR #184, opened 2025-02-20
Implement context path discovery for CalDav and CardDav endpoints

timofurrer:
See individual commits for context of the changes.

### DIFF
--- a/internal/client.go
+++ b/internal/client.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/xml"
+	"errors"
 	"fmt"
 	"io"
 	"mime"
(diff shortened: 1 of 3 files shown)


emersion:
"caldavs"/"caldavs" is not a typo. It's CalDAV/CardDAV over HTTPS. "carddav"/"caldav" are over insecure HTTP.

internal/client.go
```diff
@@ -31,23 +34,52 @@ func DiscoverContextURL(ctx context.Context, service, domain string) (string, er
...
+	}
+
+	var path string
+	switch len(txtRecords) {
+	case 0:
+		path = "/.well-known/" + service
+	case 1:
+		record := txtRecords[0]
+		if !strings.HasPrefix(record, "path=") {
+			break
```
emersion, line 63: This leaves `path` empty. Probably we should return an error instead?

emersion:
> "caldavs"/"caldavs" is not a typo. It's CalDAV/CardDAV over HTTPS. "carddav"/"caldav" are over insecure HTTP.

Oh, oops. `DiscoverContextURL` already adds the final "s". Good catch!

internal/client.go
timofurrer: Absolutely!

timofurrer:
@emersion thanks for the timely review. I've addresses your feedback and pushed an updated change set. Back to you :ping_pong:

PR #186, opened 2025-03-03
webdav: remove path from LocalFileSystem path errors

krystiancha:
We don't want to show local paths in error messages.

`os.Is*` calls were replaced with `errors.Is` to also match the new wrapped error.


emersion (approved):
Thanks!

PR #189, opened 2025-07-08
webdav: return multistatus with 403 on PROPPATCH

krystiancha:
This fixes a "The file cannot be accessed by the system" error when uploading some files with Windows Explorer.


emersion (approved):
Thanks!

PR #199, opened 2025-12-18
caldav: get calendar data using PROPFIND

voidcontext:
This change adds FindCalendar calendar to the caldav client. It can be
used to get calendar information without loading all the events once the
path of the calendar is known.

### DIFF
--- a/caldav/client.go
+++ b/caldav/client.go
@@ -3,6 +3,7 @@ package caldav
 import (
 	"bytes"
 	"context"
+	"encoding/xml"
 	"errors"
 	"fmt"
 	"mime"
@@ -16,6 +17,14 @@ import (
 	"github.com/emersion/go-webdav"
 	"github.com/emersion/go-webdav/internal"
 )
+
+var calendarProps = []xml.Name{
+	internal.ResourceTypeName,
+	internal.DisplayNameName,
+	calendarDescriptionName,
+	maxResourceSizeName,
+	supportedCalendarComponentSetName,
+}
 
 // DiscoverContextURL performs a DNS-based CardDAV service discovery as
 // described in RFC 6352 section 11. It returns the URL to the CardDAV server.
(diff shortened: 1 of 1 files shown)


emersion:
This is too specific for a particular use-case.

Could we return a full `Calendar` instead?

voidcontext:
> This is too specific for a particular use-case.
> 
> Could we return a full `Calendar` instead?

That's fair, let me have a look.

voidcontext:
@emersion I've updated the PR, this should be more general, I wasn't sure about the naming though. What do you think?

caldav/client.go
```diff
@@ -42,6 +42,31 @@ func NewClient(c webdav.HTTPClient, endpoint string) (*Client, error) {
 	return &Client{wc, ic}, nil
 }
 
+// Get calendar information without loading all the events once the path of the calendar is known.
+// As path use
+// - ${calId}/user (Google)
+// - calendar path (standard implementations)
```
emersion: Nit: can we leave out any provider-specific details here?

caldav/client.go
```diff
@@ -42,6 +42,31 @@ func NewClient(c webdav.HTTPClient, endpoint string) (*Client, error) {
...
+func (c *Client) FindCalendar(ctx context.Context, path string) (Calendar, error) {
```
emersion: Could we make this function return a `*Calendar` instead of a `Calendar`?

- We can return `nil` on error, instead of an empty struct
- The struct has quite a few fields, returning a pointer avoids copying
- A pointer is more idiomatic

caldav/client.go
```diff
@@ -42,6 +42,31 @@ func NewClient(c webdav.HTTPClient, endpoint string) (*Client, error) {
...
+	propfind := internal.NewPropNamePropFind(
```
emersion: Nit: we could extract this into a global `var` and use it in the two functions performing a `PROPFIND` on a calendar. This would ensure we always request the information nee by `decodeCalendar` from the server.

caldav/client.go
```diff
@@ -87,44 +112,53 @@ func (c *Client) FindCalendars(ctx context.Context, calendarHomeSet string) ([]C
...
-			return nil, fmt.Errorf("carddav: max-resource-size must be a positive integer")
-		}
+	return l, errors.Join(errs...)
+}
 
-		var supportedCompSet supportedCalendarComponentSet
-		if err := resp.DecodeProp(&supportedCompSet); err != nil && !internal.IsNotFound(err) {
-			return nil, err
-		}
+func decodeCalendar(path string, resp *internal.Response) (Calendar, error) {
```
emersion: Ditto: returning `*Calendar` would be nicer here

emersion (commented):
Can you edit the commit and PR title to not use conventional commits? We use the package name as the prefix, just like the Go repository does.

> I wasn't sure about the naming though. What do you think?

Right! `FindCalendar` is consistent with the existing functions, but is perhaps a bit too easy to mix up with `FindCalendars` (only one letter differs). This probably calls for a wider renaming - we can leave this up for a future issue/PR.

caldav/client.go
voidcontext: I am fairly new to go, and I still find it difficult sometimes to decide between returning a pointer or the struct. It seems I need to be more aggressive, as I wouldn't have considered the Calendar struct to have enough fields that'd warrant  a pointer. Thanks for sharing your insights!

caldav/client.go
voidcontext: Removed.

caldav/client.go
voidcontext: Actually I was thinking about doing this!

caldav/client.go
voidcontext: Refactored.

emersion (approved):
Thank you!

PR #202, opened 2026-03-23
caldav: add support for scheduling (RFC 6638)

cheif:
## Summary

This adds partial support for RFC 6638, which allows us to deliver "Inbox" events, AKA invitations to events, that the user can reply to.

I elected to implement this as a new interface, with a single requirement:
```
GetInbox(ctx context.Context) (*Inbox, error)
```

The returned inbox contains a path to the inbox, e.g. `/user/inbox`, and a list of the users addresses. When this is implemented the backend will return that it supports scheduling, and the client will then try to fetch the inbox-items the same way that it does other events, through `ListCalendarObjects`.

For the user to be able to respond to an invite, the calendar-objects returned needs to contain an `ical.PropAttendee` which should have the address as one of the addresses listed in the inbox.

When the user responds to an invitation, we'll get a `PutCalendarEvent`, just like a regular edit.

## Testing
I've verified this works as expected using the builtin apple calendar on macOS and iOS, but it would be great if someone would like to test this on other platforms as well.

### DIFF
--- a/caldav/caldav.go
+++ b/caldav/caldav.go
@@ -145,3 +145,8 @@ type SyncResponse struct {
 	Updated   []CalendarObject
 	Deleted   []string
 }
+
+type Inbox struct {
+	Path           string
+	UserAddressSet []string
+}
--- a/caldav/elements.go
+++ b/caldav/elements.go
@@ -38,11 +38,13 @@ func (a *calendarHomeSet) GetXMLName() xml.Name {
 	return calendarHomeSetName
 }
 
+// https://datatracker.ietf.org/doc/html/rfc6638#section-2.4.1
 type calendarUserAddressSet struct {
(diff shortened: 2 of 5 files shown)


emersion:
Please have a look at [CONTRIBUTING.md](https://github.com/emersion/.github/blob/main/CONTRIBUTING.md), in particular the LLM section.

cheif:
> Please have a look at [CONTRIBUTING.md](https://github.com/emersion/.github/blob/main/CONTRIBUTING.md), in particular the LLM section.

Yeah, sorry. I asked copilot with help to add some tests, and apparently it went ahead and created a PR as well 😱 

I'll rework it and get back to you, sorry for the noise!

cheif:
@emersion Any interest in getting this merged from your side?

I've been successfully running this on one of my own projects (getting the kids soccer/etc. practices into my calendar, with support for reporting their attendance, instead of using yet another app), and it seems to be working nicely for that usecase at least 😄

Issue #203, opened 2026-04-02
ResolveHref strips trailing slashes from baseURL and principal path, causing client failures


What-is-water93:
When using this library to connect to my CalDav server ([opencloud](https://github.com/opencloud-eu/opencloud)) I received a 405 when trying to get the principal via `FindCurrentUserPrincipal`.  
The issue I found via debugger was that the baseURL I set (`$mydomain.tld/caldav/`)  was changed to `$mydomain.tld/caldav`  (no trailing slash) when the request was build, resulting in a 405 response.  
This happens in the `path.Join` call of the `ResolveHref` function, which strips any `/`  from paths (outside of the path "/").
The same issue affects `FindCalendarHomeSet`, if the principal is, like in my case, something like `"/caldav/$some_id/"`, because the trailing slash is also stripped ,  which results in a 403 statuscode.  

A possible fix for this could be 9891abf9bbbaffc06f77f7983453df321559ce02.

I am not 100% sure if the behaviour of opencloud (or the radicale component they use for caldav) is correct in sending a 405/403 if the trailing slash is not set instead of a redirect, but I had no issues using the caldav endpoint before with e.g. DAVx5.

PR #204, opened 2026-04-03
internal: preserve trailing slashes in baseURL and principal path

What-is-water93:
See #203 for the problem this change would fix


emersion (approved):
Thanks!

PR #205, opened 2026-05-06
internal: tolerate unquoted ETags from non-compliant servers

jeboehm:
## Problem

RFC 4918 §15.6 requires ETags to be quoted strings, but some servers return them without surrounding double-quotes. mailbox.org CalDAV is one example:

```
<DAV: getetag>Y2FsOi8vMC8zMg</DAV: getetag>
```

`strconv.Unquote` fails with `invalid syntax` on such values, causing the whole calendar sync to abort with:

```
property <DAV: getetag>: webdav: failed to unquote ETag: invalid syntax
```

## Fix

When `strconv.Unquote` fails and the raw value is non-empty, fall back to accepting the raw value as-is. Empty strings still return an error.

## Tests

Added `TestETagUnmarshal` covering:
- Quoted ETags (RFC-compliant)
- Unquoted ETags (non-compliant server, e.g. mailbox.org)
- Quoted ETags with special characters

Added `TestETagRoundTrip` verifying that marshal→unmarshal produces the correct quoted form.

### DIFF
--- a/internal/elements.go
+++ b/internal/elements.go
@@ -385,8 +385,16 @@ type GetETag struct {
 type ETag string
 
 func (etag *ETag) UnmarshalText(b []byte) error {
-	s, err := strconv.Unquote(string(b))
+	raw := string(b)
+	s, err := strconv.Unquote(raw)
 	if err != nil {
+		// Some servers (e.g. mailbox.org) return ETags without the required
+		// surrounding double-quotes (RFC 4918 §15.6). Accept the raw value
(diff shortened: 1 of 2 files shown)


emersion (changes requested):
This is incorrect: an unquoted E-Tag will get re-quoted when marshaled, causing server-side mismatch.

emersion:
Has this been written with an LLM? See CONTRIBUTING for LLM policy.

PR #208, opened 2026-06-24
caldav, carddav: support the CalendarServer getctag property

kanocz:
## Summary

Add optional support for the `getctag` property from the CalendarServer extension
namespace (`http://calendarserver.org/ns/`). `getctag` is an opaque collection-level
token that changes whenever the collection's contents change, letting clients
(DAVx5, Apple, etc.) detect "did anything change?" with a single `PROPFIND` on the
collection instead of enumerating every member's `ETag`. It's widely supported and
is the change-detection fallback for clients that don't use `sync-collection`.

## Change

- Add a `CTag string` field to `caldav.Calendar` and `carddav.AddressBook`.
- Add the `internal.GetCTag` element, `internal.GetCTagName`, and a
  `CalendarServerNamespace` constant.
- In `propFindCalendar` / `propFindAddressBook`, report `getctag` when `CTag` is
  non-empty.

When `CTag` is empty (the default), the property is omitted entirely, so existing
backends are unaffected.

## Notes

- Although the property lives in the "calendarserver" namespace, the same
  property/namespace is also used for CardDAV address books (per the CalendarServer
  convention and as implemented by SabreDAV / dav4jvm / Apple). The shared
  `internal.GetCTag` reflects that; a clarifying comment is included in the code.

### DIFF
--- a/caldav/caldav.go
+++ b/caldav/caldav.go
@@ -69,6 +69,9 @@ type Calendar struct {
 	Description           string
 	MaxResourceSize       int64
 	SupportedComponentSet []string
+	// CTag, when set, is reported as the CalendarServer getctag property: an
+	// opaque token that must change whenever the calendar's contents change.
+	CTag string
 }
 
 type CalendarCompRequest struct {
--- a/caldav/server.go
+++ b/caldav/server.go
@@ -570,6 +570,9 @@ func (b *backend) propFindCalendar(ctx context.Context, propfind *internal.PropF
 			Size: cal.MaxResourceSize,
 		})
 	}
+	if cal.CTag != "" {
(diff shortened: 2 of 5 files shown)


emersion:
I'm not super keen about adding support for non-standard extensions. Clients should upgrade to RFC 6578.

kanocz:
It's "non standart", but de facto is... Apple introduced it and most third-party dav-clients uses it... and this allow to significantly reduce traffic and load on synchronization of big collection

mrusme:
@emersion this PR would be benefitial and it seems like @kanocz is right that many DAV clients seem to be making use of this non standard feature. Would you give it a second thought?

PR #210, opened 2026-07-23
caldav, carddav: Resolve /.well-known/

mrusme:
By default providers like Fastmail give users the root URL to their CalDAV/CardDAV servers (e.g. `https://caldav.fastmail.com`, `https://carddav.fastmail.com`) which use the `/.well-known/` route to point the clients to the correct resources. While it's possible for users to export the direct resource URL from within the web UI, it's not as user-friendly for inexperienced users.

This change should fix that by having the client look up correct URLs.


emersion:
`DiscoverContextURL` already performs this step.

`FindCurrentUserPrincipal` is not the right spot to perform auto-discovery.

mrusme:
Oh well, it seems like this is an issue with my specific DNS setup, which seemingly does not return the required DNS records properly. Thank you for clarifying.

PR #213, opened 2026-09-13
carddav: serve CS:getctag when the backend supplies one

akn101:
## Problem

A CardDAV collection served by this package advertises no change token: neither `CS:getctag` nor `DAV:sync-token` is ever reported. That is legal — RFC 6352 requires neither, and `sync-collection` (RFC 6578) is optional — but it costs every client the one optimisation the protocol offers.

Clients poll a collection with a cheap `Depth: 0` PROPFIND and skip the far more expensive `Depth: 1` enumeration for as long as the token they already hold still matches. With no token to compare against, that never engages: every poll re-reads every card.

Observed against a server built on this package, serving a 2125-contact address book. Both Apple clients do this on every cycle, for an unchanged collection:

```
macOS/27.0 AddressBookCore/2765.100.1.1.1
  PROPFIND /dav/addressbooks/<user>/<book>/   207     4610 bytes   (Depth: 0)
  PROPFIND /dav/addressbooks/<user>/<book>/   207   821246 bytes   (Depth: 1)

iOS/27.0 dataaccessd/1.0
  PROPFIND /dav/addressbooks/<user>/<book>/   207     4611 bytes   (Depth: 0)
  PROPFIND /dav/addressbooks/<user>/<book>/   207   821248 bytes   (Depth: 1)
```

~820 KB per poll, per device, to learn that nothing changed. It scales with collection size and device count.

## Change

`AddressBook` gains a `CTag` field. When non-empty it is served as `CS:getctag`; when empty nothing is fabricated and the property is reported not found exactly as today.

This is opt-in and backwards compatible: backends that set nothing are unchanged. Backends that can cheaply derive a token from their storage now have a way to expose it.

`CS:getctag` rather than `DAV:sync-token` because the latter also requires implementing the `sync-collection` REPORT, which is a much larger change to the `Backend` interface. The CTag is a small, self-contained step; adding `sync-token` later is not precluded.

## Tests

`carddav/getctag_test.go` covers both directions — served with the backend's value when set, reported not found when unset — driven through a real `PROPFIND` against `httptest`. Full suite passes, `gofmt` clean.

## Note

Verified at the protocol level: the `Depth: 0` response carries the token when the backend supplies one. I have not measured a real client skipping an enumeration against a patched server, so the traffic saving is reasoned from the protocol and the access log rather than observed.


akn101:
Correcting the original description of this PR, which overstated the case.

I first opened this having concluded that macOS Contacts *never* issues the `Depth: 1` enumeration when no change token is advertised — that it reports a successful sync of zero contacts and stays silently empty. Longer observation against the same server shows that is not true: macOS does enumerate unconditionally, exactly as iOS does. My original evidence compared two log excerpts that were, on closer inspection, against two different collections, and the empty address book I was debugging had a different cause entirely.

The doc comment on the new field made the same claim, and I have amended the commit to remove it.

The change still stands on its own: with no token to compare against, every poll re-reads the entire collection. That is a real and measurable cost — ~820 KB per poll per device on the instance above — just a performance argument rather than a correctness one. Apologies for the noise.

